PR #343, opened 2018-09-07
Compute correct transaction value for exchange/field bundles

### LINKED ISSUE #340: Incorrect transaction value for bundles created by exchanges
### Bug description
Bundles with multiple outputs leads to an incorrect calculation of transaction value. 

Related issue: https://github.com/iotaledger/trinity-wallet/issues/312 (This issue was related to only failed transactions)

### Environment Specs

- Manufacturer: Apple
- Model: 6
- OS version: 11.4
- Trinity version:  v(0.4.1) 17

### Steps To Reproduce

Pull in transactions for bundle `9RIIFXWDJAGSTXLVILRABQWIIWCQNIRSKTGCLCFQNJISXTEB9BFOKVYYEADWYBHGBLJWQJEUWAJZPKFTW` and notice transaction value

### Expected behaviour
It should calculate and display correct transaction value

### Actual behaviour
Displays incorrect transaction value

Yes.

laumair:
# Description

Fixes https://github.com/iotaledger/trinity-wallet/issues/340

## Type of change

# How Has This Been Tested?

- Added unit test coverage

# Checklist:

### DIFF
--- a/src/shared/__tests__/libs/iota/transfers.spec.js
+++ b/src/shared/__tests__/libs/iota/transfers.spec.js
@@ -43,30 +43,127 @@ describe('libs: iota/transfers', () => {
     });
 
     describe('#getTransferValue', () => {
+        let ownAddresses;
+
+        before(() => {
+            ownAddresses = ['A'.repeat(81), 'B'.repeat(81), 'C'.repeat(81), 'D'.repeat(81), 'E'.repeat(81)];
+        });
+
         describe('zero value transactions', () => {
             it('should return zero', () => {
                 // Zero value transactions have no inputs
                 expect(
                     getTransferValue(
                         [],
-                        [{ value: 0, currentIndex: 0, lastIndex: 1 }, { value: 0, currentIndex: 1, lastIndex: 1 }],
+                        [
+                            {
+                                value: 0,
+                                currentIndex: 0,
+                                lastIndex: 1,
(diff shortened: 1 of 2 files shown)


src/shared/__tests__/libs/iota/transfers.spec.js
```diff
@@ -43,30 +43,127 @@ describe('libs: iota/transfers', () => {
...
+            });
+
+            describe('with no input addresses belong to user addresses', () => {
+                it('should return a sum of all user output addresses', () => {
+                    expect(
+                        getTransferValue(
+                            [
+                                {
+                                    value: -10,
+                                    currentIndex: 0,
```
rajivshah3, line 132: Wouldn't the transaction with a `currentIndex` of 0 be an output/have a positive value?

laumair: I don't think it's necessary but I'll sort the indexes. Good catch.

rajivshah3: Yeah, I didn't think it would affect anything really

rajivshah3: Actually this isn't a big deal, we can merge this anyway

rajivshah3 (approved):

PR #344, opened 2018-09-07
Ensure react-native-view-shot saves to internal cache

cvarley100:
# Description

react-native-view-shot saves to external cache. With sharing QRs this introduces the risk of a man-in-the-disk attack see: https://blog.checkpoint.com/2018/08/12/man-in-the-disk-a-new-attack-surface-for-android-apps/ Attacker can swap shared QR for one encoding their own address.

## Type of change

# How Has This Been Tested?

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration_

Tested on Android/iOS

# Checklist:


src/mobile/package.json
```diff
@@ -101,7 +101,7 @@
     "react-native-triangle": "0.0.8",
     "react-native-udp": "^1.2.9",
     "react-native-vector-icons": "^4.5.0",
-    "react-native-view-shot": "^2.3.0",
+    "react-native-view-shot": "git+https://github.com/cvarley100/react-native-view-shot#master",
```
rajivshah3, line 104: Does the `yarn.lock` need to be updated as well?

rajivshah3 (approved):

PR #350, opened 2018-09-09
Desktop: Update in memory seed usage

rihardsgravis:
# Description

- Add missing manual Garbage Collector trigger after seed usage
- Switch to byte array for all code base

## Type of change

# How Has This Been Tested?

Tested on macOS using `lldb`

# Checklist:


src/desktop/src/ui/views/account/Tools.js
```diff
@@ -51,13 +51,23 @@ class Addresses extends PureComponent {
...
+    componentDidUpdate(prevProps) {
+        const { wallet, ui } = this.props;
+
+        if (
+            prevProps.isTransitioning === ui.isTransitioning &&
+            prevProps.isAttachingToTangle === ui.isAttachingToTangle &&
+            prevProps.balanceCheckFlag === wallet.balanceCheckFlag &&
+            prevProps.ui.isSyncing === ui.isSyncing
+        ) {
+            return false;
```
laumair: Is there a reason for returning a boolean in `componentDidUpdate`?

rihardsgravis: No, will update

src/desktop/src/libs/helpers.js
```diff
@@ -13,6 +13,68 @@ export const capitalize = (input) => {
...
+ */
+const charToByte = (char) => {
+    return '9ABCDEFGHIJKLMNOPQRSTUVWXYZ'.indexOf(char);
+};
+
+const byteToTrit = (byte) => {
+    return trytesTrits[byte % 27];
+};
+
+module.exports = {
```
rajivshah3, line 74: Was there an issue with exporting the constant itself?

src/desktop/src/ui/views/wallet/Receive.js
```diff
@@ -128,7 +128,10 @@ class Receive extends React.PureComponent {
                     >
                         <p>
                             {receiveAddress.split('').map((char, index) => {
-                                const scrambleChar = scramble[index] > 0 ? byteToChar(scramble[index]) : null;
+                                const scrambleChar =
+                                    scramble[index] > 0
+                                        ? '9ABCDEFGHIJKLMNOPQRSTUVWXYZ'.charAt(scramble[index] % 27)
```
rajivshah3: Might be easier if we have this charset stored in another file

rajivshah3 (approved):

src/desktop/src/ui/views/wallet/Receive.js
rihardsgravis: We actually, do it's `charToByte` function at `libs/helpers`. Will update to use that.

src/desktop/src/libs/helpers.js
rihardsgravis: `libs/helpers` is now used also on the main process and that is not going through Babel, so no ES6, here :/

PR #434, opened 2018-09-29
Add local timezone support

### LINKED ISSUE #338: Use system timezone for timestamps
### Description
Timestamps currently use UTC instead of the local timezone for the system. `moment-timezone` should be implemented in order to show the timestamps in the local timezone.

### Motivation
- It's not apparent to users that the timestamp is in UTC
- It's easier for users to understand timestamps in their own timezone rather than UTC

### Requirements

- [x] Detect system timezone
- [x] Change timestamps to system timezone

Yes

rajivshah3:
# Description
Shows timestamps in local timezone instead of UTC

Closes #338 

## Type of change

# How Has This Been Tested?

- [x] Tested manually on desktop
- [x] Tested manually on iOS

# Checklist:

### DIFF
--- a/src/desktop/src/ui/components/Chart.js
+++ b/src/desktop/src/ui/components/Chart.js
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import { LineChart, ResponsiveContainer, Line, YAxis, Tooltip } from 'recharts';
-import { format } from 'date-fns';
+import { formatTimeAs, detectedTimezone } from 'libs/date';
 
 import withChartData from 'containers/components/Chart';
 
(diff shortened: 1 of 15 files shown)


src/desktop/src/ui/components/Chart.js
```diff
@@ -54,7 +54,7 @@ class Chart extends PureComponent {
             const date = new Date(props.payload[0].payload.time * 1000);
             return (
                 <p className={css.label}>
-                    {format(date, 'DD.MM.YY HH:mm')}
+                    {formatTimeAs.dayMonthYearHoursMinutes(navigator.language, detectedTimezone, date)}
```
rihardsgravis, line 57: Should we use system language (`navigator.language`) or the user chosen wallet language 🤔

rihardsgravis (approved):

src/desktop/src/ui/components/Chart.js
rajivshah3: I used the system language so that the functionality would be the same as mobile (https://github.com/iotaledger/trinity-wallet/pull/434/files#diff-cd6bfd84c7bace320f0bf28eab3ba245R8). `moment` supports more locales than we do, so if the user uses English because their native language isn't supported by Trinity, it will still use their native date/time format to match how the rest of their device will format dates/times.

cvarley100 (approved):

PR #435, opened 2018-09-29
Desktop: Use a separate keychain entry for development versions

rajivshah3:
# Description
Keeps the keychains for the development and production versions of Trinity separate. This ensures that if a developer is running a development version of Trinity, it will never have access to the seed used in the production version (unless the developer intentionally uses the same seed in both versions). 

## Type of change

# How Has This Been Tested?

- [x] Tested manually on macOS

# Checklist:

### DIFF
--- a/src/desktop/lib/preload/Electron.js
+++ b/src/desktop/lib/preload/Electron.js
@@ -44,6 +44,9 @@ let locales = {
 let onboardingSeed = null;
 let onboardingGenerated = false;
 
+// Use a different keychain entry for development versions
+const KEYTAR_SERVICE = currentWindow.isPackaged ? 'Trinity wallet' : 'Trinity wallet (dev)';
+
 /**
  * Global Electron helper for native support
  */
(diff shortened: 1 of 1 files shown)


src/desktop/lib/preload/Electron.js
```diff
@@ -44,6 +44,9 @@ let locales = {
 let onboardingSeed = null;
 let onboardingGenerated = false;
 
+// Use a different keychain entry for development versions
+const KEYTAR_SERVICE = currentWindow.isPackaged ? 'Trinity wallet' : 'Trinity wallet (dev)';
```
rihardsgravis: `isPackaged` is available only starting from Electron `3.0` (this check would always return dev now as `isPackaged` is undefined). To update to that we need to update the binary package builds to support node `10.2.0`. I have this in my next weeks todo.

This  merge should then wait till then, or we could change the check to `process.env.NODE_ENV === "development" ? 'Trinity wallet (dev)' : 'Trinity wallet'`

rihardsgravis (changes requested):

src/desktop/lib/preload/Electron.js
rajivshah3: Ah, I didn't realize that. I've changed to `process.env` in df850f3

rihardsgravis (approved):

PR #380, opened 2018-09-17
Seed storage modules

rihardsgravis:
# Description

Decouple seed storing, address generation and signing from state actions to separate modules.

## Type of change

- Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?

Tested on macOS, Ubuntu, iOS.

# Checklist:

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] For changes to `mobile` that include native code (including React Native modules): I have verified that both iOS and Android successfully build in both `Debug` and `Release` modes
- [x] For changes to `shared`: If applicable, I have verified that my changes are implemented correctly in `desktop` and `mobile`

PR #480, opened 2018-10-08
Update input selection process

laumair:
# Description

Currently, inputs are selected in a sequential way (starting from the least index with balance and stopping at the index where it satisfies the provided threshold). 

Input selection now accepts a _limit_ to restrict the number of inputs wherever necessary as is expected to work as follows:

**Case 1: When no limit is provided**
Select inputs by optimal value i.e. select inputs as few as possible. 

**Case 2: When limit is provided**
Try to select inputs by optimal value i.e. select inputs as few as possible. If selected inputs exceed limit then: 

If possible, select a set of inputs within range where _Sum_(inputs) == transaction value. This will avoid remainders. Otherwise, select a set of least size inputs within range.

Todo: Need to fix/update some tests related to [makeTransaction](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/actions/transfers.js#L414)

Depends on https://github.com/iotaledger/trinity-wallet/pull/380

## Type of change

- Breaking change 

# How Has This Been Tested?

- Added unit test coverage

# Checklist:


rajivshah3 (approved):

src/desktop/src/libs/SeedStore/Keychain.js
```diff
@@ -11,13 +11,17 @@ class Keychain {
      * Init the vault
      * @param {array} key - Account decryption key
      * @param {string} accountId - Account identifier
+     * @param {number} maxInputs - Maximum number of inputs allowed per bundle
      */
-    constructor(key, accountId) {
+    constructor(key, accountId, maxInputs) {
```
rihardsgravis: I would suggest this to move as a static read-only property of the seedvault class itself (as it should only change form one seedvault type to other), see proposed codebase in the Ledger implementation.

laumair: Resolved in cc51b20

PR #399, opened 2018-09-19
Upgrade nav lib and refactor modal system

cvarley100:
# Description

Updates RNN to v2 
Closes https://github.com/iotaledger/trinity-wallet/issues/383.

## Type of change

- Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?

# Checklist:

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] For changes to `mobile` that include native code (including React Native modules): I have verified that both iOS and Android successfully build in both `Debug` and `Release` modes
- [x] For changes to `shared`: If applicable, I have verified that my changes are implemented correctly in `desktop` and `mobile`

PR #483, opened 2018-10-09
Purge persisted state on wallet reset

laumair:
# Description

Purging stored state didn't remove persisted state from storage because of a wrong `persister` import. 
This fixes persistor's import and uses an alternative way of purging the stored state. Also, removes unnecessarily dispatched state actions for resetting state values.

## Type of change

- Bug fix

# How Has This Been Tested?

- Observing _persisted_ state after wallet reset

# Checklist:


laumair:
@rajivshah3 
CI is failing with 
```
Cannot find entry file index.android.js in any of the roots: ["/Users/vagrant/git/src/mobile","/Users/vagrant/git/src/shared"]
```

I think this is related to the changes to entry point files in the [base](https://github.com/iotaledger/trinity-wallet/tree/feat/global-modal) branch. 

I think we don't remove cached node modules in the PR workflow of bitrise. (Adding that to the workflow) should fix this problem.

cvarley100 (approved):

rajivshah3:
@laumair the CI is failing because of the final step which bundles the JS code. The command it runs is `react-native bundle --entry-file index.android.js --platform android --bundle-output android/main.jsbundle --dev true` but in #399 the `--entry-file` was changed to index.js. I can fix this with a script that checks first if `index.js` exists so this doesn't happen

cvarley100 (approved):

PR #535, opened 2018-10-31
#470 Feat/rename seedvault file

### LINKED ISSUE #470: Rename seedvault file
### Description
The exported seedvault file should by default be named like the account it was exported from. For example: mainaccount20180905.kdbx

### Environment Specs
Trinity Desktop and Trinity mobile

### Motivation
its more userfriendly

no

delki8:
# Description
Changing the prefix of the suggested (on desktop) exported filename to match the account name being exported. If for some reason we export more than one seed, the standard 'seedvault' prefix is applied.

Fixes #470 

## Type of change

# How Has This Been Tested?
I exported a SeedVault file for my personal account and check that the suggested filename would contain the account name.

# Checklist:


src/desktop/lib/preload/Electron.js
```diff
@@ -366,10 +366,10 @@ const Electron = {
         try {
             const content = await kdbx.exportVault(seeds, password);
             const now = new Date();
-
+            const prefix = seeds.length === 1 ? seeds[0].title : 'seedvault';
```
rajivshah3:
```suggestion
            const prefix = seeds.length === 1 ? seeds[0].title : 'SeedVault';
```
I'm being really picky here 😄  would you mind making this change so that we can have consistency with mobile?

rajivshah3 (dismissed):

src/desktop/lib/preload/Electron.js
delki8: No problem at all

src/desktop/lib/preload/Electron.js
```diff
@@ -366,10 +366,10 @@ const Electron = {
         try {
             const content = await kdbx.exportVault(seeds, password);
             const now = new Date();
-
+            const prefix = seeds.length === 1 ? seeds[0].title : 'SeedVault';
```
rihardsgravis: We should probably do a sanity check for special characters (eg `./\`), UTF-16 symbols (eg emojis) that could make the file name "troublesome" on some platforms. Maybe just leave ASCII symbols and filter out special characters?

delki8: Great point, I'll address this in my next commit.

delki8: @rajivshah3 and @rihardsgravis I couldn't find a better way to do it than duplicating the function inside _desktop_ and _shared_ folders, because it isn't possible to require a _shared/utils.js_ function from within a file that is not compiled through webpack. Is there a nicer alternative?

rihardsgravis: That is fine for now! I added an issue #548 to add the `imports` support for the main process scripts.

But in addition of removing non ASCII characters we should also remove all special characters like `/ \ : * ? " < > |` and spaces. Or maybe it would be easier to just leave alpha numeric values and `-`?

delki8: Agreed, alphanumeric with `-` will be easier.

src/mobile/src/ui/components/SeedVaultExportComponent.js
```diff
@@ -20,6 +20,7 @@ import InfoBox from './InfoBox';
 import Button from './Button';
 import CustomTextInput from './CustomTextInput';
 import PasswordFields from './PasswordFields';
+import { removeNonAlphaNumeric } from '../../../../shared/libs/utils';
```
rajivshah3:
```suggestion
import { removeNonAlphaNumeric } from 'shared-modules/libs/utils';
```
We recently added https://github.com/tleunen/babel-plugin-module-resolver so now the `../..` is no longer needed. If you want, you can change this, but it's not a huge issue.

src/mobile/src/ui/components/SeedVaultExportComponent.js
```diff
@@ -20,6 +20,7 @@ import InfoBox from './InfoBox';
 import Button from './Button';
 import CustomTextInput from './CustomTextInput';
 import PasswordFields from './PasswordFields';
+import { removeNonAlphaNumeric } from 'shared-modules/libs/utils';
```
rajivshah3: The CI is failing with:
> $ ./node_modules/.bin/eslint src/mobile/
/Users/vagrant/git/src/mobile/src/ui/components/SeedVaultExportComponent.js
  23:39  error  Absolute imports should come before relative imports  import/first
✖ 1 problem (1 error, 0 warnings)

My bad, I should have told you beforehand. To make ESLint happy, you'll have to move this above the relative imports:

```js
import { isAndroid, getAndroidFileSystemPermissions } from 'libs/device';
import { removeNonAlphaNumeric } from 'shared-modules/libs/utils'; <--
import InfoBox from './InfoBox';
import Button from './Button';
import CustomTextInput from './CustomTextInput';
import PasswordFields from './PasswordFields';
```

Once that's done, we should be good to go

delki8: kkk Ok, don't worry about it.

rajivshah3 (approved):
Looks good! Thanks @delki8

rihardsgravis (approved):

PR #575, opened 2018-11-05
Desktop: Update css and highlight receive address checksum

cvarley100:
# Description

- Highlight checksum for receive address
- Update border radius

## Type of change

# Checklist:

### DIFF
--- a/src/desktop/src/ui/views/wallet/Receive.js
+++ b/src/desktop/src/ui/views/wallet/Receive.js
@@ -204,7 +204,7 @@ class Receive extends React.PureComponent {
                                     .substring(81, 90)
                                     .split('')
                                     .map((char, index) => {
-                                        const scrambleChar = scramble[index] > 0 ? byteToChar(scramble[index]) : null;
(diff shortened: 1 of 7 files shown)


src/desktop/src/ui/views/wallet/Receive.js
```diff
@@ -190,10 +190,28 @@ class Receive extends React.PureComponent {
...
+                                    return (
+                                        <React.Fragment key={`char-${index}`}>{scrambleChar || char}</React.Fragment>
+                                    );
+                                })}
+                            <span>
+                                {receiveAddress
+                                    .substring(81, 90)
+                                    .split('')
+                                    .map((char, index) => {
+                                        const scrambleChar = scramble[index] > 0 ? byteToChar(scramble[index]) : null;
```
rihardsgravis: The `index` is reset, so the animation would be more correct like:
```suggestion
                                        const scrambleChar = scramble[index + 81] > 0 ? byteToChar(scramble[index + 81]) : null;
```

rihardsgravis (approved):

PR #598, opened 2018-11-13
Fix condition on spend status in libs/iota/removeUnusedAddresses

laumair:
# Description

`removeUnusedAddresses` is used in initial account sync and manual sync. Consider a case where an address has no transactions, no balance and spent status true. The incorrect condition will remove the address and wouldn't display it to the user until `syncAccount` is called (e.g., login, before transactions). 

Tests for `removeUnusedAddresses` were all passing because there was no test coverage added for such a case (It was only for positive balances and transaction hashes).

This PR fixes the condition and also adds coverage for the above mentioned case.

Note: This can only be reproduced on a fresh install of wallet because manual sync does not remove seen addresses from wallet. 

## Type of change

- Bug fix 

# How Has This Been Tested?

- Manually tested by using a seed with an address with no transaction hashes, no balance and spent status true
- Added unit test coverage for the above mentioned scenario. 

# Checklist:

### DIFF
--- a/src/shared/__tests__/libs/iota/addresses.spec.js
+++ b/src/shared/__tests__/libs/iota/addresses.spec.js
@@ -1014,25 +1014,70 @@ describe('libs: iota/addresses', () => {
             sandbox.restore();
         });
 
-        describe('when the last address has associated meta data', () => {
-            it('should return addresses with latest unused address', () => {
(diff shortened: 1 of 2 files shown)


rajivshah3 (approved):
Looks good! 👍

PR #541, opened 2018-10-31
Desktop:  Disable Remote debugging port on production build

rihardsgravis:
# Description

Disable Remote debugging port on production build

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

# How Has This Been Tested?

Tested manually on macOS, Windows and Ubuntu

# Checklist:

_Please delete items that are not relevant._

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

PR #610, opened 2018-11-19
Desktop: Terminate application if Remote Debugging Port is detected

### LINKED ISSUE #604: DevTools with websocket access always running by default (appimage)
Starting the .appimage via command line on linux shows 
> DevTools listening on ws://127.0.0.1:35437/devtools/browser/998eaa1d-9840-4d08-850d-449efaa073ad

This address seems to be a websocket and accept json. not tested further. however, now its possible to access http://127.0.0.1:35437/json and get more information about trinity. especially the 

> "devtoolsFrontendUrl": "/devtools/inspector.html?ws=127.0.0.1:35437/devtools/page/E6558DE92883401BA3D32963E229AB01"
> 
which rewritten like this allows opening trinity inside the browser and use the browsers own dev tools / console

http://127.0.0.1:35437/devtools/inspector.html?ws=127.0.0.1:35437/devtools/page/E6558DE92883401BA3D32963E229AB01

I don't know if that poses a potential security risk, but at least i think devtools should not running by default / there should be no open websocket that could allow access.

- OS: 4.19.2-1-MANJARO
- Trinity version: 0.4.0

rihardsgravis:
# Description

Unsetting `remote-debugging-port` (#541) on runtime results in running the debug mode on random port rather disabling it. The application can only be terminated if debug mode is detected.

Fixes #604

## Type of change

### DIFF
--- a/src/desktop/main.js
+++ b/src/desktop/main.js
@@ -12,15 +12,10 @@ const electronSettings = require('electron-settings');
 app.commandLine.appendSwitch('js-flags', '--expose-gc');
 
 /**
- * Disable Remote Debugging Port on production build
- */
-app.commandLine.appendSwitch('remote-debugging-port', '');
-
-/**
  * Terminate application if Node remote debugging detected
  */
 const argv = process.argv.join();
-if (argv.includes('inspect') || typeof v8debug !== 'undefined') {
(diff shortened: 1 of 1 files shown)


rajivshah3 (approved):
Works great, tested on macOS 👍

Issue #453, opened 2018-10-03
Node quorum


laumair:
### Description
Trinity only connects to a single node for fetching transactions, latest confirmation states, address spend statuses and latest balances. Trinity does make sure that the node it connects to is [synced](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/libs/iota/extendedApi.js#L510) but this does not guarantee whether the information it's getting from the node is correct. 

 In a distributed computing environment, a single source of information is ultimately fallible. The implication of this — when applied to a wallet — is a limit to the level of confidence we can have in the responses of a single node. This is where node quorum becomes important. Confidence increases when responses of multiple nodes are considered in tandem. With 2/3 of a pool of nodes in agreement, we can be confident that the information returned is correct. And in the case where the nodes fail to agree, we fallback to a safe result.  

### Motivation

To ensure Trinity always pulls in correct information (transactions, balances, spend statuses, confirmation states)

### Requirements

- Consider all addresses `spent` unless majority of the queried nodes say otherwise.
- Consider balance `zero` unless majority of the queried nodes say otherwise.
- Consider transaction as `unconfirmed` unless majority of the queried  nodes say otherwise. 

For new transactions, we should find unique transaction hashes from all nodes. Then compute a quorum result for valid trytes. 
### Am I planning to do it myself with a PR?
Yes

ovanwijk:
Even though this code is also to ensure a quick response it does show an interesting way for dealing with multiple nodes. :https://github.com/ovanwijk/IotaPublicNodeSelection/blob/695d9b91056ebcc4d5aa27d50e7024823f399877/src/MultiNodeInterface.js#L594

What it does is basically select a set of nodes, query them all at the same time. If it gets the first result back all other nodes have to reply within 2 seconds (increase it to 30 seconds to negate the wait and have a quorum). In case of wereAddressesSpentFrom it treats any TRUE as more important then a set of FALSE.

laumair:
This is indeed an interesting way. 

> In case of wereAddressesSpentFrom it treats any TRUE as more important then a set of FALSE.

One problem that I see is that a _bad_ node could block funds. However, this approach could be well suited for `findTransactions` as hashes & signatures can be validated.

laumair:
Done in https://github.com/iotaledger/trinity-wallet/pull/631

Also related https://github.com/iotaledger/trinity-wallet/pull/884

PR #631, opened 2018-11-22
Node Quorum

laumair:
# Description

Fixes # https://github.com/iotaledger/trinity-wallet/issues/453

Depends on https://github.com/iotaledger/iota.js/pull/314

- `wereAddressesSpentFrom` - Compute a quorum result from multiple nodes. Fallback to `true` if there is no consensus among results. 
- `getInclusionStates ` - First compute a quorum result for `latestSolidSubtangleMilestone`, then compute a quorum for inclusion states by using the agreed `latestSolidSubtangleMilestone` as a tip. Fallback to `false` if there is no consensus among results.
- `getBalances` - Similar to `getInclusionStates`. First compute a quorum result for `latestSolidSubtangleMilestone`, then compute a quorum for balances by using the agreed `latestSolidSubtangleMilestone` as a tip. Fallback to `'0'` if there is no consensus among results.
- ~~`getTrytes` - Compute a quorum result from multiple nodes. Fallback to null hash trytes if there is no consensus among results.~~ 
- ~~`findTransactions` - The strategy for determining a quorum result for this endpoint is different than the rest. Computing a quorum for`findTransactions` simply prepare a set of unique hashes from multiple nodes.~~
## Type of change

- Enhancement

# How Has This Been Tested?

**TODO**

# Checklist:


src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+
+    return Promise.resolve();
+};
+
+/**
+ *   Determines quorum result for supported methods.
+ *
+ *   @method determineQuorumResult
+ *   @param {array} validResults
+ *   @param {number} faultyResultCount
```
alon-e: I think this changed in the impl (to quorumSize), but not the JSDocs.

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+ *
+ *   @returns {function}
+ **/
+const determineQuorumResult = (validResults, quorumSize = QUORUM_SIZE) => {
+    const { frequency, mostFrequent } = findMostFrequent(validResults);
+
+    return (method, threshold) => {
+        // Include faulty result count while determining percentage
+        const percentage = frequency[mostFrequent] / quorumSize * 100;
```
alon-e, line 75: even though it evaluates correctly. I think this is clearer:
```suggestion
        const percentage = (frequency[mostFrequent] / quorumSize) * 100;
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+/**
+ *   Determines quorum result for supported methods.
+ *
+ *   @method determineQuorumResult
+ *   @param {array} validResults
+ *   @param {number} faultyResultCount
+ *
+ *   @returns {function}
+ **/
+const determineQuorumResult = (validResults, quorumSize = QUORUM_SIZE) => {
```
alon-e, line 70: how are valid results defined? (what is an invalid result, if someone did not answer?)

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+ *   @param {array} [blacklistedNodes = []]
+ *
+ *   @returns {Promise}
+ **/
+const findSyncedNodes = (nodes, quorumSize, syncedNodes = [], blacklistedNodes = []) => {
+    const sizeOfSyncedNodes = size(syncedNodes);
+    const whitelistedNodes = filter(nodes, (node) => !includes(blacklistedNodes, node) && !includes(syncedNodes, node));
+
+    if (isEmpty(whitelistedNodes)) {
+        return Promise.reject(new Error(Errors.NOT_ENOUGH_HEALTHY_QUORUM_NODES));
```
alon-e: could it be that case that `syncedNodes` is large enough -> `|syncedNodes| = quorumSize`, and this is the reason that `isEmpty(whitelistedNodes)` ?

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+/**
+ *   From a list of synced nodes, compute a quorum result for wereAddressesSpentFrom iota api
+ *
+ *   @method getQuorumForWereAddressesSpentFrom
+ *   @param {array} payload - Addresses
+ *   @param {array} syncedNodes
+ *
+ *   @returns {Promise}
+ **/
+const getQuorumForWereAddressesSpentFrom = (payload, syncedNodes) => {
```
alon-e: the following 4 (?) functions are duplicates except for the method name (which is used to invoke the API and get the safe default). 

could this be done using the `method` as an argument? and have only 1 copy of this function, the reason being that this mechanism is new and might need to be tweaked, maintaining all the copies is a hassle and would lead to errors.


and later adding more methods to quorum would be easier as well.

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+        )
+        .then((results) => {
+            const validResults = filter(results, (result) => !isUndefined(result));
+
+            let idx = 0;
+            const quorumResult = [];
+
+            while (idx < requestPayloadSize) {
+                /* eslint-disable no-loop-func */
+                quorumResult.push(
```
alon-e: IIUC, if the request is for L addresses the Quorom is determined for each address separately, correct?
if so, that's great! (what I envisioned) but it should be clearly stated (in the JSDocs?)

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+    const requestPayloadSize = size(payload);
+
+    return rejectIfNotEnoughSyncedNodes(syncedNodes)
+        .then(() =>
+            Promise.all(
+                map(
+                    syncedNodes,
+                    (provider) =>
+                        new Promise((resolve) => {
+                            new IOTA({ provider }).api.wereAddressesSpentFrom(
```
alon-e: does this have some type of timeout? so err (or valid response) is promised to be returned eventually?

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+         * Performs a quorum for getLatestInclusion api endpoint
+         *
+         * @method getLatestInclusion
+         * @param {array} hashes
+         *
+         * @returns {Promise}
+         */
+        getLatestInclusion(hashes) {
+            return findSyncedNodesIfNecessary().then((newSyncedNodes) =>
+                getQuorumForLatestSolidSubtangleMilestone(newSyncedNodes).then((latestSolidSubtangleMilestone) =>
```
alon-e: this addition is worth mentioning somewhere - in the JSDocs (same for `getBalances`)

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+                            );
+                        }),
+                ),
+            ),
+        )
+        .then((results) => {
+            const validResults = filter(results, (result) => !isUndefined(result));
+
+            // Get quorum result for latest solid subtangle milestone
+            const latestSolidSubtangleMilestone = determineQuorumResult(
```
alon-e: nit-pick, but this should follow the structure of the code above, it makes maintenance harder (I understand that there is no payload and the response is a single value)

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,415 @@
...
+                map(validResults, (nodeInfo) => nodeInfo.latestSolidSubtangleMilestone),
+            )('latestSolidSubtangleMilestone', QUORUM_THRESHOLD);
+
+            // If nodes cannot agree on the latestSolidSubtangleMilestone
+            // Then just throw an exception
+            if (latestSolidSubtangleMilestone === EMPTY_HASH_TRYTES) {
+                // TODO (laumair): Might need to replace the error message here.
```
alon-e: good idea :)

alon-e (changes requested):
looks great!
requested changes mostly b/c comment doesn't affect review count.

my main open question is if there is a timeout for unresponsive nodes (to mark invalid).

laumair:
@alon-e Thank you for your review. 

> my main open question is if there is a timeout for unresponsive nodes (to mark invalid).

This was indeed missing. Have fixed it [here](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L272). [getIotaInstance](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/extendedApi.js#L31) sets the default timeout.

src/shared/libs/iota/quorum.js
laumair: Good catch. [Fixed](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L31).

src/shared/libs/iota/quorum.js
laumair: That's correct. Invalid results are basically the results from unresponsive nodes. 

If there's a network error, we explicitly mark the result as `undefined` [here](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L273) and later on filter these results.

src/shared/libs/iota/quorum.js
laumair: That was actually a minor bug. Thank you. 

The only reason why this should reject is if there no whitelisted nodes and we still need to fill up _synced_ nodes to reach the quorum size. Have fixed it [here](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L127) .

src/shared/libs/iota/quorum.js
laumair: Great suggestion. 

One reason why I avoided doing that in the first place is because some methods (e.g. getBalances, getNodeInfo) response is not a plain array of booleans like the rest of the methods (wereAddressesSpentFrom, getLatestInclusion). However, it's not a big problem. I agree that all four methods had duplication and maintaining all four would be a hassle. 

Have simplified the implementation. Now, there's only a single [function](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L261) that serves all four endpoints and also a separate [function](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L180) for preparing results. 

Reason why I decoupled preparation of results from `#getQuorum` is because it would be a lot easier to unit test both separately.

src/shared/libs/iota/quorum.js
laumair: That's correct. Quorum is determined for each address separately. 
Have added better code comments and also have updated [JSDocs](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L167-L171).

src/shared/libs/iota/quorum.js
laumair: It will eventually timeout (because of the addition [here)](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L272) and an error or a valid response is guaranteed.

src/shared/libs/iota/quorum.js
laumair: Did you mean documentation for first getting quorum results for `latestSolidSubtangleMilestone`? Have updated [JSDocs](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L345-L347) for this method also for [getBalances](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L368-L370).

src/shared/libs/iota/quorum.js
laumair: Good suggestion. As mentioned [here](https://github.com/iotaledger/trinity-wallet/pull/631#discussion_r241632229), the pattern is now consistent with these [changes](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L180-L284).

src/shared/libs/iota/quorum.js
laumair: Also added [test coverage](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/__tests__/libs/iota/quorum.spec.js#L144) for this.

src/shared/libs/iota/quorum.js
alon-e: To clarify that this is intentional (seems good to me)
the quorum is calculated *only* within the responsive nodes.
this should be mentioned somewhere - b/c it changes the strategy to attack the quorum.

for example:
```
N = 7
m = 3 (malicious)
scenario A:
all m vote No
all honest vote Yes
Result: 4/7= FailSafe

scenario B:
2 m vote "unresponsive"
1 m votes No
all honest vote Yes
Result: 3/4= Yes
```

so the best stratagy is to participate.
OTOH,
if the malicious nodes manage to kick off honest nodes (the will become uresponsive), they can pass bad elections:
```

scenario C:
3 m vote No
2 honest vote Yes
2 honest vote "unresponsive" due to malice
Result: 3/5= No
(well that's cutting it close, but you get the drift)
```

to give a full picture - anouther option would be to have "unresponsive" vote FailSafe.
but propbably needs some analysis as well.

laumair: > the quorum is calculated only within the responsive nodes.

The quorum is not calculated within the responsive nodes. Unresponsive nodes would also be considered. Explicitly passing `undefined` as a result for unresponsive nodes is just to easily filter out invalid results. 

Quorum [calculation](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L73) would always be out of `N` (`N === quorumSize`). 

> scenario B:
2 m vote "unresponsive"
1 m votes No
all honest vote Yes
Result: 3/4= Yes

In this case, result would be 3/N. 

> anouther option would be to have "unresponsive" vote FailSafe.

IIUC, then always calculating quorum out of `N` basically does the same?

alon-e (approved):

src/shared/libs/iota/quorum.js
alon-e: >The quorum is not calculated within the responsive nodes. Unresponsive nodes would also be considered. Explicitly passing undefined as a result for unresponsive nodes is just to easily filter out invalid results.

oh, so that is confusing -

> That's correct. Invalid results are basically the results from unresponsive nodes.

 the param `validResults` sounds like it's only the.. well ... valid results. filtering the unresponsive nodes.

alon-e: oh, so `quorumSize = QUORUM_SIZE` is `N`, regardless of the size of `validResults` - :+1: 
maybe worth another pass to make it a bit clearer :)

laumair: > oh, so quorumSize = QUORUM_SIZE is N, regardless of the size of validResults

That's correct. I'll make another commit to make it clearer. Sorry about the confusion.

laumair: [This](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d0a4dd9c0a12e24610e6238b13de71fce19a8543) should improve the readability of the overall implementation.

src/shared/config.js
```diff
@@ -44,3 +44,7 @@ export const NODE_REQUEST_TIMEOUT = 6000 * 2;
 export const DEFAULT_RETRIES = 4;
 
 export const IRI_API_VERSION = '1';
+
+export const QUORUM_THRESHOLD = 67;
+export const QUORUM_NODES_SIZE = 7;
```
cvarley100: I think we can change the variable names here
```suggestion
export const QUORUM_SIZE = 7;
```

src/shared/config.js
```diff
@@ -44,3 +44,7 @@ export const NODE_REQUEST_TIMEOUT = 6000 * 2;
...
+export const QUORUM_NODES_SYNC_CHECKS_INTERVAL = 120;
```
cvarley100:
```suggestion
export const QUORUM_SYNC_CHECK_INTERVAL = 120;
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+import size from 'lodash/size';
+import split from 'lodash/split';
+import sampleSize from 'lodash/sampleSize';
+import union from 'lodash/union';
+import uniq from 'lodash/uniq';
+import { isNodeHealthy, getIotaInstance } from './extendedApi';
+import {
+    QUORUM_THRESHOLD,
+    QUORUM_NODES_SIZE,
+    QUORUM_NODES_SYNC_CHECKS_INTERVAL,
```
cvarley100:
```suggestion
    QUORUM_SYNC_CHECK_INTERVAL,
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+import transform from 'lodash/transform';
+import size from 'lodash/size';
+import split from 'lodash/split';
+import sampleSize from 'lodash/sampleSize';
+import union from 'lodash/union';
+import uniq from 'lodash/uniq';
+import { isNodeHealthy, getIotaInstance } from './extendedApi';
+import {
+    QUORUM_THRESHOLD,
+    QUORUM_NODES_SIZE,
```
cvarley100:
```suggestion
    QUORUM_SIZE,
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+/**
+ * Resolves only if size of synced nodes is greater than or equal to quorum size
+ *
+ * @method rejectIfNotEnoughSyncedNodes
+ *
+ * @param {array} nodes
+ * @param {number} quorumSize
+ * @returns {Promise}
+ */
+const rejectIfNotEnoughSyncedNodes = (nodes, quorumSize = QUORUM_NODES_SIZE) => {
```
cvarley100:
```suggestion
const rejectIfNotEnoughSyncedNodes = (nodes, quorumSize = QUORUM_SIZE) => {
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+        throw new Error(Errors.NOT_ENOUGH_QUORUM_NODES);
+    }
+
+    let syncedNodes = [];
+    let lastSyncedAt = new Date();
+
+    const findSyncedNodesIfNecessary = () => {
+        const timeElapsed = (new Date() - lastSyncedAt) / 1000;
+
+        if (isEmpty(syncedNodes) || timeElapsed >= QUORUM_NODES_SYNC_CHECKS_INTERVAL) {
```
cvarley100:
```suggestion
        if (isEmpty(selectedNodes) || timeElapsed >= QUORUM_SYNC_CHECK_INTERVAL) {
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+/**
+ * Resolves only if provided milestone is not equal to empty hash trytes (999...999)
+ *
+ * @method rejectIfEmptyHashTrytes
+ *
+ * @param {string} latestSolidSubtangleMilestone
+ *
+ * @returns {Promise<string>}
+ */
+const rejectIfEmptyHashTrytes = (latestSolidSubtangleMilestone) => {
```
cvarley100: This is a weird method name

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+/**
+ *   Determines quorum result for supported methods.
+ *
+ *   @method determineQuorumResult
+ *   @param {array} validResults - Results from responsive nodes
+ *   @param {number} quorumSize - Size of quorum nodes
+ *
+ *   @returns {function(string, [number]): {boolean | string}}
+ **/
+const determineQuorumResult = (validResults, quorumSize = QUORUM_NODES_SIZE) => {
```
cvarley100:
```suggestion
const determineQuorumResult = (validResults, quorumSize = QUORUM_SIZE) => {
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+
+    if (!includes(keys(allowedMethodsMap), method)) {
+        throw new Error(Errors.METHOD_NOT_SUPPORTED_FOR_QUORUM);
+    }
+
+    return allowedMethodsMap[method];
+};
+
+/**
+ *   For a list of nodes, find healthy (synced) nodes N, where size(N) === QUORUM_NODES_SIZE
```
cvarley100:
```suggestion
 *   For a list of nodes, find healthy (synced) nodes N, where size(N) === QUORUM_SIZE
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+ *
+ *   @method Quorum
+ *   @param {array} quorumNodes
+ *
+ *   @returns {object}
+ **/
+export default function Quorum(quorumNodes) {
+    let nodes = uniq(quorumNodes);
+
+    if (size(nodes) < QUORUM_NODES_SIZE) {
```
cvarley100:
```suggestion
    if (size(nodes) < QUORUM_SIZE) {
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    let syncedNodes = [];
+    let lastSyncedAt = new Date();
+
+    const findSyncedNodesIfNecessary = () => {
+        const timeElapsed = (new Date() - lastSyncedAt) / 1000;
+
+        if (isEmpty(syncedNodes) || timeElapsed >= QUORUM_NODES_SYNC_CHECKS_INTERVAL) {
+            return findSyncedNodes(nodes, QUORUM_NODES_SIZE, syncedNodes).then((newSyncedNodes) => {
```
cvarley100:
```suggestion
            return findSyncedNodes(nodes, QUORUM_SIZE, selectedNodes).then((syncedNodes) => {
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+ *   @param {number} quorumSize - Size of quorum nodes
+ *
+ *   @returns {function(string, [number]): {boolean | string}}
+ **/
+const determineQuorumResult = (validResults, quorumSize = QUORUM_NODES_SIZE) => {
+    const { frequency, mostFrequent } = findMostFrequent(validResults);
+
+    return (method, threshold = QUORUM_THRESHOLD) => {
+        // Always calculate percentage out of quorum size i.e., all (responsive + unresponsive) nodes.
+        const percentage = frequency[mostFrequent] / quorumSize * 100;
```
cvarley100, line 75: We could remove `100` and just do quorum threshold as a decimal

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+const fallbackToSafeResult = (method) => {
+    const allowedMethodsMap = {
+        wereAddressesSpentFrom: true,
+        getInclusionStates: false,
+        'getBalances:balances': '0',
+        'getNodeInfo:latestSolidSubtangleMilestone': EMPTY_HASH_TRYTES,
+    };
+
+    if (!includes(keys(allowedMethodsMap), method)) {
+        throw new Error(Errors.METHOD_NOT_SUPPORTED_FOR_QUORUM);
```
cvarley100, line 102: 👍

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+ *   @returns {object}
+ **/
+export default function Quorum(quorumNodes) {
+    let nodes = uniq(quorumNodes);
+
+    if (size(nodes) < QUORUM_NODES_SIZE) {
+        throw new Error(Errors.NOT_ENOUGH_QUORUM_NODES);
+    }
+
+    let syncedNodes = [];
```
cvarley100:
```suggestion
    let selectedNodes = [];
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+    let lastSyncedAt = new Date();
+
+    const findSyncedNodesIfNecessary = () => {
+        const timeElapsed = (new Date() - lastSyncedAt) / 1000;
+
+        if (isEmpty(syncedNodes) || timeElapsed >= QUORUM_NODES_SYNC_CHECKS_INTERVAL) {
+            return findSyncedNodes(nodes, QUORUM_NODES_SIZE, syncedNodes).then((newSyncedNodes) => {
+                syncedNodes = newSyncedNodes;
```
cvarley100:
```suggestion
                selectedNodes = syncedNodes;
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+                lastSyncedAt = new Date();
+
+                return newSyncedNodes;
```
cvarley100:
```suggestion
                return syncedNodes;
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+            });
+        }
+
+        return Promise.resolve(syncedNodes);
```
cvarley100:
```suggestion
        return Promise.resolve(selectedNodes);
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+         *
+         * @method wereAddressesSpentFrom
+         * @param {array} addresses
+         *
+         * @returns {Promise}
+         */
+        wereAddressesSpentFrom(addresses) {
+            return isEmpty(addresses)
+                ? Promise.resolve([])
+                : findSyncedNodesIfNecessary().then((newSyncedNodes) =>
```
cvarley100:
```suggestion
                : findSyncedNodesIfNecessary().then((syncedNodes) =>
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+                      getQuorum('wereAddressesSpentFrom', newSyncedNodes, addresses),
```
cvarley100:
```suggestion
                      getQuorum('wereAddressesSpentFrom', syncedNodes, addresses),
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+         *
+         * @method getLatestInclusion
+         * @param {array} hashes
+         *
+         * @returns {Promise}
+         */
+        getLatestInclusion(hashes) {
+            return isEmpty(hashes)
+                ? Promise.resolve([])
+                : findSyncedNodesIfNecessary().then((newSyncedNodes) =>
```
cvarley100:
```suggestion
                : findSyncedNodesIfNecessary().then((syncedNodes) =>
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+                      getQuorum('getNodeInfo:latestSolidSubtangleMilestone', newSyncedNodes)
+                          // If nodes cannot agree on the latestSolidSubtangleMilestone
+                          // No need to proceed further.
+                          .then(rejectIfEmptyHashTrytes)
+                          .then((latestSolidSubtangleMilestone) =>
+                              getQuorum('getInclusionStates', newSyncedNodes, hashes, [latestSolidSubtangleMilestone]),
```
cvarley100:
```suggestion
                              getQuorum('getInclusionStates', syncedNodes, hashes, [latestSolidSubtangleMilestone]),
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+         * @param {array} addresses
+         * @param {number} [threshold]
+         *
+         * @returns {Promise}
+         */
+        getBalances(addresses, threshold = DEFAULT_BALANCES_THRESHOLD) {
+            return isEmpty(addresses)
+                ? Promise.resolve([])
+                : findSyncedNodesIfNecessary().then((newSyncedNodes) =>
+                      getQuorum('getNodeInfo:latestSolidSubtangleMilestone', newSyncedNodes)
```
cvarley100:
```suggestion
                      getQuorum('getNodeInfo:latestSolidSubtangleMilestone', syncedNodes)
```

src/shared/libs/iota/quorum.js
```diff
@@ -0,0 +1,407 @@
...
+                          // If nodes cannot agree on the latestSolidSubtangleMilestone
+                          // No need to proceed further.
+                          .then(rejectIfEmptyHashTrytes)
+                          .then((latestSolidSubtangleMilestone) =>
+                              getQuorum('getBalances:balances', newSyncedNodes, addresses, threshold, [
```
cvarley100:
```suggestion
                              getQuorum('getBalances:balances', syncedNodes, addresses, threshold, [
```

cvarley100 (changes requested):

src/shared/libs/iota/quorum.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR396)

src/shared/libs/iota/quorum.js
laumair: Resolved [here](https://github.com/laumair/trinity-wallet/blob/feat/quorum/src/shared/libs/iota/quorum.js#L372)

src/shared/libs/iota/quorum.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR310)

src/shared/libs/iota/quorum.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR29)

src/shared/libs/iota/quorum.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR15)

src/shared/config.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR15)

src/shared/config.js
laumair: Resolved [here](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR15)

src/shared/libs/iota/quorum.js
laumair: It rejects a promise if provided trytes are null hash trytes. Have changed the [paramter](https://github.com/iotaledger/trinity-wallet/pull/631/commits/d8ad41e6a738ac82b528e3fc7ef87ee6225b956b#diff-78157a2166aa6eaa7e812e543c8729ddR46) to just `trytes` for clarity

cvarley100 (approved):

PR #692, opened 2018-12-04
Desktop: Check for a unique seed for keychain type accounts

### LINKED ISSUE #681: Desktop - We can add the same seed many times
### Bug description
The same seed can be added many times on Trinity Desktop

### Environment Specs
PC - Windows 7

- Trinity version: 
0.4.3 

### Steps To Reproduce
1. Create account with your certain seed
2. Create another account with the same seed
3. Create another account with the same seed
4. Repeat forever because Trinity Desktop allows you :D **(on Trinity Mobile, this isn't allowed and we're receiving an allert)**

### Expected behaviour
We should receive an alert like in Trinity Mobile when we're trying to add the same seed again.

No

rihardsgravis:
# Description

Add unique seed check for keychain type accounts

Fixes #681

## Type of change

# Checklist:

### DIFF
--- a/src/desktop/src/libs/SeedStore/Keychain.js
+++ b/src/desktop/src/libs/SeedStore/Keychain.js
@@ -196,6 +196,40 @@ class Keychain {
     };
 
     /**
+     * Unique seed check
+     * @param {array} Seed - Seed to check
+     * @returns {boolean} If Seed is unique
+     */
+    isUniqueSeed = async (seed) => {
+        const vault = await Electron.listKeychain();
+        if (!vault) {
+            throw new Error('Local storage not available');
+        }
+        try {
+            const accounts = vault.filter((acc) => acc.account !== ACC_MAIN && acc.account !== `${ACC_MAIN}-salt`);
+
+            for (let i = 0; i < accounts.length; i++) {
+                const account = accounts[i];
+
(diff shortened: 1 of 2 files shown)


src/desktop/src/libs/SeedStore/Keychain.js
```diff
@@ -195,6 +195,40 @@ class Keychain {
...
+        try {
+            const accounts = vault.filter((acc) => acc.account !== ACC_MAIN && acc.account !== `${ACC_MAIN}-salt`);
+
+            for (let i = 0; i < accounts.length; i++) {
+                const account = accounts[i];
+
+                const vaultSeed = await decrypt(account.password, this.key);
+
+                if (
+                    vaultSeed.length === seed.length &&
```
rajivshah3, line 217: Doesn't this check make this one redundant? https://github.com/iotaledger/trinity-wallet/blob/f050c68d1d44ab1ec3adaa50f8b2310ce197124b/src/desktop/src/ui/views/onboarding/SeedVerify.js#L86

rihardsgravis: The seeds checked against could not be the same length as the input seed (in theory it should, but you never know).

rajivshah3 (approved):

PR #777, opened 2018-12-12
Finalise Realm instantiation 

laumair:
# Description

Finalise the way realm is initialized on desktop, mobile & test environments. 

## Type of change

- Bug fix

# How Has This Been Tested?

Tested by building desktop & mobile and running tests for both `shared` and `mobile` directories. 

# Checklist:

### DIFF
--- a/src/desktop/native/Realm.js
+++ b/src/desktop/native/Realm.js
@@ -1,0 +1,7 @@
+const Realm = require('realm');
+
+// Don't report Realm JS analytics data
+// @see https://github.com/realm/realm-js/blob/master/lib/submit-analytics.js#L28
+process.env.REALM_DISABLE_ANALYTICS = 'true';
+
+module.exports = Realm;
--- a/src/desktop/native/preload/Electron.js
+++ b/src/desktop/native/preload/Electron.js
@@ -13,6 +13,7 @@ const kdbx = require('../kdbx');
 const Entangled = require('../Entangled');
 const { byteToTrit, byteToChar, removeNonAlphaNumeric } = require('../../src/libs/helpers');
 const ledger = require('../hardware/Ledger');
+const Realm = require('../Realm');
 
 const capitalize = (string) => {
     return string.charAt(0).toUpperCase() + string.slice(1).toLowerCase();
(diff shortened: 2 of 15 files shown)


rajivshah3 (approved):

rajivshah3 (commented):
On src/mobile/package.json, `REALM_DISABLE_ANALYTICS=true` needs to be added to the beginning of the `yarn test` command otherwise Buildkite may hang until the build times out

laumair:
@rajivshah3 Perhaps this should be automatically set from the Bitrise (via environment variable) for all mobile & desktop builds?

rajivshah3:
@laumair I think it might be best to specify it in the `test` command. While it doesn't hang when run locally or on Bitrise, the console output of most tests cases looks like this:

image

On Buildkite, the build will hang until it times out (up to 3 hours): https://buildkite.com/iota-foundation/trinity-mobile-prs/builds/1198#f1213926-ed40-4b07-a6f4-adcc0157e478

If we add `REALM_DISABLE_ANALYTICS=true`, we can mitigate these with a single change instead of specifying it in multiple CI configurations

laumair:
@rajivshah3 Cool have added https://github.com/iotaledger/trinity-wallet/pull/777/commits/c031b607196013a4298b72bc6d3276fb33da5a63

I suppose this should also be added to production builds. We should be able to do it directly via Bitrise for prod builds.

rajivshah3:
@laumair good point. Are we able to disable it in-app as well? If we can't, we may have to update the privacy policy

laumair:
> Are we able to disable it in-app as well?

I don't follow, sorry. What do you mean by _in-app_?

rajivshah3:
@laumair I think Realm submits analytics back to their developers unless the analytics are disabled. I'm  wondering if these analytics would also be sent while an end user is running Trinity

https://github.com/realm/realm-js/blob/master/lib/submit-analytics.js

laumair:
I believe that they would be sent unless they are disabled. Personally, I don't see any reason why we should enable them in prod.

rajivshah3:
Yeah, I don't either. I think it would require us to update the privacy policy anyway. Is there a way we can set `process.env` in React Native? It looks like Rihards already added it for desktop: https://github.com/iotaledger/trinity-wallet/pull/777/files#diff-76257644db42119f694e9a6149c87360R5

rajivshah3:
@laumair looks like it actually doesn't run in production, so we should be good: https://github.com/realm/realm-js/blob/master/react-native/ios/RealmReact/RealmAnalytics.mm#L23

PR #834, opened 2018-12-20
Mobile Alpha 0.6.0 (33)

cvarley100:


src/mobile/src/ui/components/AnimatedComponent.js
```diff
@@ -145,6 +145,10 @@ class AnimatedComponent extends Component {
                 case 'slideOutRightSmall':
                     animatedStyle = merge({}, animatedStyle, { width, transform: [{ translateX: this.slideValue }] });
                     break;
+                case 'slideInBottom':
+                case 'slideOutBottom':
+                    animatedStyle = merge({}, animatedStyle, { width, transform: [{ translateY: this.slideValue }] });
```
laumair, line 162: Since `animatedStyle` is defined as inside this function and as a new object a deep merge can be avoided here and we can just reassign to `animatedStyle`

src/mobile/src/ui/components/TopBar.js
```diff
@@ -233,6 +232,22 @@ class TopBar extends Component {
         }
     }
 
+    /**
+     * Returns padding styling dependent on device type
+     *
+     * @method getTopbarPadding
+     * @returns {object}
```
laumair: The return type should be of type `any` since there's also a possibility that this function could return an `undefined` value (In case of iPhoneX).

src/mobile/src/ui/views/wallet/Receive.js
```diff
@@ -748,6 +804,7 @@ const mapStateToProps = (state) => ({
     usdPrice: state.marketData.usdPrice,
     conversionRate: state.settings.conversionRate,
     hadErrorGeneratingNewAddress: state.ui.hadErrorGeneratingNewAddress,
+    ui: state.ui,
```
laumair: Misses `propType` validation.

laumair (changes requested):

PR #868, opened 2018-12-31
Check if address is valid

### LINKED ISSUE #866: No error with invalid addresses
### Bug description
No error if you want to send value to an invalid address (last trit is not 0)

### Environment Specs
- OS version: Win 10

What build of Trinity are you running? Please provide the full version (you can find it in Settings > About Trinity).

- Trinity version: 0.4.6

### Steps To Reproduce
1. Try to send value to an address of which the last trit is not 0 (-1 or 1)

### Expected behaviour
An error should be returned stating that the address is invalid

### Actual behaviour
The bundle will be completed and an error is returned when sending to the node

### Errors
`Request error: Invalid trytes input`

7h7:
# Description

Check for value transactions that the last trit from the address is 0, otherwise it's invalid and the bundle isn't accepted by the node
https://github.com/iotaledger/iri/blob/1ec6dcc007a0cd119d6a9c9878842515390ced9b/src/main/java/com/iota/iri/TransactionValidator.java#L179-L181

Fixes #866

## Type of change

# Checklist:

### DIFF
--- a/src/shared/__tests__/libs/iota/utils.spec.js
+++ b/src/shared/__tests__/libs/iota/utils.spec.js
@@ -19,7 +19,7 @@ describe('libs: iota/utils', () => {
         });
         describe('when the last trit is not 0', () => {
             it('should return false', () => {
-                expect(isLastTritZero('E'.repeat(81)).to.equal(false);
+                expect(isLastTritZero('E'.repeat(81))).to.equal(false);
             });
         });
     });
--- a/src/shared/__tests__/libs/iota/utils.spec.js
+++ b/src/shared/__tests__/libs/iota/utils.spec.js
@@ -7,6 +7,7 @@ import {
     getRandomNodes,
     withRetriesOnDifferentNodes,
     throwIfNodeNotHealthy,
+    isLastTritZero,
 } from '../../../libs/iota/utils';
 
 describe('libs: iota/utils', () => {
--- a/src/shared/__tests__/libs/iota/utils.spec.js
+++ b/src/shared/__tests__/libs/iota/utils.spec.js
@@ -10,6 +10,18 @@ import {
 } from '../../../libs/iota/utils';
 
 describe('libs: iota/utils', () => {
+    describe('#isLastTritZero', () => {
+        describe('when the last trit is 0', () => {
+            it('when the last trit is 0', () => {
+                expect(isLastTritZero('D'.repeat(81)).to.equal(true);
+            });
+        });
+        describe('when the last trit is not 0', () => {
+            it('should return false', () => {
+                expect(isLastTritZero('E'.repeat(81)).to.equal(false);
+            });
+        });
+    });
     describe('#convertFromTrytes', () => {
         describe('when trytes passed as an argument contains all nines', () => {
             it('should return a string "Empty"', () => {
--- a/src/shared/__tests__/libs/iota/utils.spec.js
+++ b/src/shared/__tests__/libs/iota/utils.spec.js
@@ -14,7 +14,7 @@ describe('libs: iota/utils', () => {
     describe('#isLastTritZero', () => {
         describe('when the last trit is 0', () => {
             it('should return true', () => {
-                expect(isLastTritZero('D'.repeat(81)).to.equal(true);
+                expect(isLastTritZero('D'.repeat(81))).to.equal(true);
             });
         });
         describe('when the last trit is not 0', () => {
--- a/src/shared/__tests__/libs/iota/utils.spec.js
+++ b/src/shared/__tests__/libs/iota/utils.spec.js
@@ -12,7 +12,7 @@ import {
 describe('libs: iota/utils', () => {
     describe('#isLastTritZero', () => {
         describe('when the last trit is 0', () => {
-            it('when the last trit is 0', () => {
+            it('should return true', () => {
                 expect(isLastTritZero('D'.repeat(81)).to.equal(true);
             });
         });
--- a/src/shared/actions/transfers.js
+++ b/src/shared/actions/transfers.js
@@ -433,7 +433,7 @@ export const makeTransaction = (seedStore, receiveAddress, value, message, accou
     const withPreTransactionSecurityChecks = () => {
         // Progressbar step => (Validating receive address)
         dispatch(setNextStepAsActive());
-        //Check the last trit for validity
+        // Check the last trit for validity
         if(isLastTritZero(address) === false){
           throw new Error(Errors.INVALID_LAST_TRIT);
         }
--- a/src/shared/actions/transfers.js
+++ b/src/shared/actions/transfers.js
@@ -30,6 +30,7 @@ import {
     fetchRemoteNodes,
     getRandomNodes,
     throwIfNodeNotHealthy,
+    isLastTritZeroAsync,
 } from '../libs/iota/utils';
 import { setNextStepAsActive, reset as resetProgress } from './progress';
 import { clearSendFields } from './ui';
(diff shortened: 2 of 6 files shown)


src/shared/__tests__/libs/iota/utils.spec.js
```diff
@@ -10,6 +10,18 @@ import {
 } from '../../../libs/iota/utils';
 
 describe('libs: iota/utils', () => {
+    describe('#isLastTritZero', () => {
+        describe('when the last trit is 0', () => {
+            it('should return true', () => {
+                expect(isLastTritZero('D'.repeat(81)).to.equal(true);
```
rajivshah3:
```suggestion
                expect(isLastTritZero('D'.repeat(81))).to.equal(true);
```

src/shared/__tests__/libs/iota/utils.spec.js
```diff
@@ -10,6 +10,18 @@ import {
...
+            });
+        });
+        describe('when the last trit is not 0', () => {
+            it('should return false', () => {
+                expect(isLastTritZero('E'.repeat(81)).to.equal(false);
```
rajivshah3:
```suggestion
                expect(isLastTritZero('E'.repeat(81))).to.equal(false);
```

src/shared/actions/transfers.js
```diff
@@ -433,7 +433,10 @@ export const makeTransaction = (seedStore, receiveAddress, value, message, accou
     const withPreTransactionSecurityChecks = () => {
         // Progressbar step => (Validating receive address)
         dispatch(setNextStepAsActive());
-
+        //Check the last trit for validity
```
rajivshah3: Small stylistic change
```suggestion
        // Check the last trit for validity
```

src/shared/libs/iota/utils.js
```diff
@@ -191,6 +191,21 @@ export const isValidAddress = (address) => {
...
+/**
+ * Checks if the last trit is 0
+ *
+ * @method isLastTritZero
+ * @param {string} address
+ *
+ * @returns {boolean}
+ */
+export const isLastTritZero = (address) => {
+    if (!/[E-V]/.test(address.slice(80,81)) {
```
rajivshah3:
```suggestion
    if (!/[E-V]/.test(address.slice(80, 81))) {
```

src/shared/actions/transfers.js
```diff
@@ -433,7 +433,10 @@ export const makeTransaction = (seedStore, receiveAddress, value, message, accou
...
+        if(isLastTritZero(address) === false){
```
rajivshah3: Small stylistic change
```suggestion
        if (isLastTritZero(address) === false){
```

src/shared/__tests__/libs/iota/utils.spec.js
```diff
@@ -10,6 +10,18 @@ import {
 } from '../../../libs/iota/utils';
```
rajivshah3, line 11: `isLastTritZero` needs to be imported
```suggestion
    isLastTritZero,
} from '../../../libs/iota/utils';
```

src/shared/actions/transfers.js
rajivshah3: `isLastTritZero` needs to be imported [here](https://github.com/7h7/trinity-wallet/blob/develop/src/shared/actions/transfers.js#L28) as well (it won't let me comment on that line):
```js
import {
    withRetriesOnDifferentNodes,
    fetchRemoteNodes,
    getRandomNodes,
    throwIfNodeNotHealthy,
    isLastTritZero,
} from '../libs/iota/utils';
```

rajivshah3 (changes requested):
Thanks for the PR @7h7 ! The following changes should make the CI pass

laumair:
Under which circumstances would Trinity generate an address with a non-zero trit? 

I think this validation should also be added to [isBundle](https://github.com/iotaledger/iota.js/blob/master/lib/utils/utils.js#L388) so that these bundles are never added to Trinity.

rajivshah3:
> Under which circumstances would Trinity generate an address with a non-zero trit?

I don’t think these addresses can be generated with a lib. They are normally human-generated. An example @7h7 gave was 
https://thetangle.org/address/TROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLT. You can try to send a value transfer to this address, but IRI will reject it. The problem is, at that point the transaction is signed and the signature is exposed to the node. 

> I think this validation should also be added to isBundle so that these bundles are never added to Trinity.

I agree

laumair:
> I don’t think these addresses can be generated with a lib. They are normally human-generated. An example @7h7 gave was
https://thetangle.org/address/TROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLTROLLT. You can try to send a value transfer to this address, but IRI will reject it. The problem is, at that point the transaction is signed and the signature is exposed to the node.

Okay thanks for the explanation. This check should be added for recipient addresses.

rajivshah3 (approved):
Looks good to me! Thanks @7h7

src/shared/libs/iota/utils.js
```diff
@@ -191,6 +191,36 @@ export const isValidAddress = (address) => {
...
+
+/**
+ * Async version of isLastTritZero
+ *
+ * @method isLastTritZeroAsync
+ * @param  {string}  address
+ *
+ * @returns {boolean}
+ */
+export const isLastTritZeroAsync = (address) => {
```
laumair: I think adding an async util here is unnecessary. 

I can see the motivation for its addition and [usage](https://github.com/iotaledger/trinity-wallet/pull/868/files#diff-8ea5b2a5a0c274aaec60a3bbffa271c0R436) but you could also do:

```
return Promise.resolve(isLastTritZero(...)).then(() => {}).catch(() => {})
```

src/shared/libs/iota/utils.js
```diff
@@ -191,6 +191,36 @@ export const isValidAddress = (address) => {
...
+/**
+ * Checks if the last trit is 0
+ *
+ * @method isLastTritZero
+ * @param {string} address
+ *
+ * @returns {boolean}
+ */
+export const isLastTritZero = (address) => {
+    if (!/[E-V]/.test(address.slice(80, 81))) {
```
laumair: Just a suggestion but if you're not doing any validation on the parameter perhaps this can be simplified to just `export const isLastTritZero = (address) => !/[E-V]/.test(address.slice(80, 81))` i.e., no need for explicit returns.

laumair (changes requested):

src/shared/actions/transfers.js
```diff
@@ -427,124 +428,132 @@ export const makeTransaction = (seedStore, receiveAddress, value, message, accou
...
-                dispatch(setNextStepAsActive());
-
-                return syncAccount()(accountState, seedStore);
-            })
-            .then((newState) => {
-                // Assign latest account but do not update the local store yet.
-                // Only update the local store with updated account information after this transaction is successfully completed.
-                accountState = newState;
+    // Check the last trit for validity
+    if (!lastTritIsZero(address)) {
```
rajivshah3: We want this to be part of the promise chain (since there's a possibility that we throw an error here, we want it to throw and then go to the `catch` block and show the error message to the user), so I think the best approach would be similar to what we had before (but incorporate @laumair 's suggestions):
```js
return Promise.resolve(isLastTritZero(address)).then((lastTritIsZero) => {
  if(!lastTritIsZero) {
    throw new Error(Errors.INVALID_LAST_TRIT);
  }
  
  return isAnyAddressSpent()([address])
  ...
```
The reason why we use these [promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises) is twofold: it makes error handling a bit easier, while also allowing us to use asynchronous functions without our code looking like [this](http://callbackhell.com/) :trollface: . So this is essentially what we're doing:

1. Call `isLastTritZero`, passing `address` as a parameter
2. Store whatever `isLastTritZero` returns in a variable called `lastTritIsZero` (in our case, this will be either true or false)
3. Is the value of `lastTritIsZero` ["falsy"](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)? (so in our case, we're essentially checking if `lastTritIsZero === false`)
 a. i. If so, throw an error
 a. ii. When the error is thrown, we'll skip all the way to the `catch` block so we can figure out what to do with the error https://github.com/iotaledger/trinity-wallet/blob/252a1ed3de6f453beef52f6c510c376979990e3f/src/shared/actions/transfers.js#L726
  b. i. If not, then we'll continue by calling `isAnyAddressSpent`

laumair (approved):
@7h7 @rajivshah3  💯

PR #877, opened 2019-01-03
Desktop: Realm Database - remove redux persist settings dependency

rihardsgravis:
# Description

- Update tray application state synchronisation and initialisation
- Update Proxy settings to use separate electronSettings entry

## Type of change

### DIFF
--- a/src/desktop/main.js
+++ b/src/desktop/main.js
@@ -227,18 +227,17 @@ function createWindow() {
             } catch (error) {}
         }
     });
-
-    if (process.platform === 'darwin') {
-        const enabled = settings ? settings.isTrayEnabled : true;
-        setupTray(enabled);
-    }
 }
 
 /**
  * Setup Tray icon
  * @param {boolean} enabled - determine if tray is enabled
  */
-const setupTray = (enabled) => {
+ipc.on('tray.enable', (e, enabled) => {
+    if (process.platform !== 'darwin') {
+        return;
+    }
+
     if (enabled === false) {
         if (tray && !tray.isDestroyed()) {
             tray.destroy();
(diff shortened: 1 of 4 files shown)


src/desktop/main.js
```diff
@@ -73,22 +71,22 @@ let windowState = {
...
-    app.commandLine.appendSwitch('auto-detect', 'false');
-    app.commandLine.appendSwitch('no-proxy-server');
-}
+try {
+    const ignoreProxy = electronSettings.get('ignore-proxy');
+    if (ignoreProxy) {
+        app.commandLine.appendSwitch('auto-detect', 'false');
+        app.commandLine.appendSwitch('no-proxy-server');
+    }
+} catch (error) {}
```
cvarley100, line 89: Do we need to do anything on catch?

rihardsgravis: Can't think of anything we could do. This can fail if the settings cannot be read from the disk, nothing much we can do in that case.

src/desktop/src/index.js
```diff
@@ -79,6 +79,18 @@ new Promise((resolve, reject) => {
         // Assign accountIndex to every account in accountInfo if it is not assigned already
         store.dispatch(assignAccountIndexIfNecessary(get(data, 'accounts.accountInfo')));
 
+        if (Electron.mode === 'tray') {
+            // Add Realm change listener to sync read-only Tray app with Main app
+            realm.addListener('change', () => {
```
laumair, line 84: Nice!

laumair:
LGTM! 💯

src/desktop/src/index.js
rajivshah3: Do we need to remove this listener at some point, or is it automatically removed when the app quits?

rihardsgravis: It's removed when the app quits and should be always active while the app runs.

rajivshah3 (approved):

PR #878, opened 2019-01-03
Shared: Fix periodically failing sortTransactionTrytesArray test

rajivshah3:
# Description

Fixes `sortTransactionTrytesArray` test that occasionally failed because it relied on `shuffle` actually changing the order of the trytes. This PR ensures that they are always in a different order and not equal to the original array.

## Type of change

# How Has This Been Tested?

- Ran tests many times without failure

# Checklist:

### DIFF
--- a/src/shared/__tests__/libs/iota/transfers.spec.js
+++ b/src/shared/__tests__/libs/iota/transfers.spec.js
@@ -4,6 +4,7 @@ import find from 'lodash/find';
 import keys from 'lodash/keys';
 import map from 'lodash/map';
 import shuffle from 'lodash/shuffle';
+import isEqual from 'lodash/isEqual';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import nock from 'nock';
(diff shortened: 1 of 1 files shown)


cvarley100 (approved):

src/shared/__tests__/libs/iota/transfers.spec.js
```diff
@@ -846,11 +847,20 @@ describe('libs: iota/transfers', () => {
...
     describe('#sortTransactionTrytesArray', () => {
         it('should sort transaction trytes in ascending order', () => {
             // failedTrytesWithCorrectTransactionHashes is in ascending order by default
-            const trytes = shuffle(failedTrytesWithCorrectTransactionHashes);
+            let shuffled = false;
+            let trytes = null;
+
+            // Ensures `trytes` never deeply equals `failedTrytesWithCorrectTransactionHashes`
+            // so that this test does not randomly fail
+            while (!shuffled) {
```
laumair: I think we should try to avoid conditionals & loops in unit tests because they become hard to read and maintain. Each condition represent a test scenario of it's own. 

For this case specifically, I think using a `shuffled` list as an input in the first place was a bad idea. IMO, this should be simplified. 

I suggest just having the input list in opposite order and then asserting if it was sorted correctly. We can just [reverse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) the input trytes.

laumair: For example:

```
it('should sort transaction trytes in ascending order', () => {
            // failedTrytesWithCorrectTransactionHashes is in ascending order by default
            const trytes = failedTrytesWithCorrectTransactionHashes.slice().reverse(); -> Changed this.
            const result = sortTransactionTrytesArray(trytes, 'currentIndex', 'asc');

           assert(result);
});
```

laumair (changes requested):

src/shared/__tests__/libs/iota/transfers.spec.js
rajivshah3: That’s a good point. I can fix this tomorrow when I get some time

rajivshah3: Fixed in a4824bd

laumair (approved):

PR #885, opened 2019-01-07
Desktop: Update garbage collector usage

rihardsgravis:
# Description

Update garbage collector usage

## Type of change

# How Has This Been Tested?

Manually on macOS using `lldb`

# Checklist:

### DIFF
--- a/src/desktop/src/libs/SeedStore/Keychain.js
+++ b/src/desktop/src/libs/SeedStore/Keychain.js
@@ -153,6 +153,8 @@ class Keychain {
             seed[i % seed.length] = 0;
         }
 
+        Electron.garbageCollect();
+
         return addresses;
     };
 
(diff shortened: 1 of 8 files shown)


src/desktop/src/libs/SeedStore/Keychain.js
```diff
@@ -169,7 +171,15 @@ class Keychain {
      */
     prepareTransfers = async (transfers, options = null) => {
         const seed = await this.getSeed(true);
-        return prepareTransfersAsync()(seed, transfers, options);
+        const transfer = await prepareTransfersAsync()(seed, transfers, options);
+
+        for (let i = 0; i < seed.length * 3; i++) {
+            seed[i % seed.length] = 0;
+        }
```
rajivshah3, line 178: We might want to use something more efficient here, such as `seed.length = 0` (see https://stackoverflow.com/a/1232046/3188334 for more info and other solutions). Just a suggestion though; I don't think a `for` loop should have too much impact

rajivshah3 (approved):
Looks good! Just one note:

cvarley100:
@rihardsgravis Does GC have any obvious performance consequences on desktop?

PR #955, opened 2019-01-18
Mobile: Update gradle, biometric error messages and fix modal issue

cvarley100:
# Description

- Adds more verbose Biometric error messages
- Fixes biometric modal issue

Fixes # (issue)

## Type of change

- Updates gradle

# How Has This Been Tested?

Tested on device

# Checklist:


laumair:
Why are we removing `src/mobile/android/app/src/main/assets/index.android.bundle`?

cvarley100:
It shouldn't be committed in the first place. We need to add it to gitignore.

cvarley100:
> Why are we removing `src/mobile/android/app/src/main/assets/index.android.bundle`?

It shouldn't be committed in the first place. We need to add it to gitignore.

rajivshah3:
I added `index.android.bundle` to the gitignore in #968. It might cause a merge conflict here, though

laumair:
@cvarley100 Merge conflicts.

rajivshah3 (approved):

src/mobile/src/ui/components/EnterPassword.js
```diff
@@ -103,7 +105,9 @@ class EnterPassword extends Component {
     }
 
     hideModal() {
-        this.props.toggleModalActivity();
+        if (this.props.isModalActive) {
+            this.props.toggleModalActivity();
```
laumair: What happens if `toggleModalActivity` is dispatched when modal is already inactive?

cvarley100: Well this function is just for hiding the modal.

laumair: Oh didn't realise it was for just hiding modal as it also [accepts](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/actions/ui.js#L217) `modalProps` and `modalContent`. 

Perhaps the condition (if modal is active) is moved to the redux action itself. 

```
const hideModal = () => {
    return (dispatch, getState) => {
        if (getState().ui.isModalActive) { dispatch(toggleModalActivity()); };
    };
};
```

That way we won't need to include `isModalActive` prop in every component where we need to close the modal.

cvarley100: This was more as a failsafe. I don't think they are necessary however - I will remove them.

cvarley100: FYI, `toggleModalActivity` has two purposes - opening the right modal with the right props. But without props it simply closes an open modal.

PR #994, opened 2019-01-28
[WIP] Desktop: Add Unit Tests

rihardsgravis:
# Description

Add Unit Tests for:

- [x] React UI components
- [x] React UI views visual
- [x] Native library wrappers
- [x] SeedStore classes
- [x] End-to-end tests for:
  - [x] Onboarding existing seed 
  - [x] Onboarding new seed
  - [ ] Send & receive

## Type of change

# Checklist:


src/desktop/package.json
```diff
@@ -51,13 +54,18 @@
     "electron-builder": "^20.38",
     "electron-devtools-installer": "^2.2.4",
     "electron-screenshot": "^1.0.5",
+    "enzyme": "^3.8.0",
+    "enzyme-adapter-react-16": "^1.8.0",
+    "enzyme-to-json": "^3.3.5",
     "express": "^4.16.3",
     "file-loader": "^2.0.0",
     "html-webpack-plugin": "^3.2.0",
+    "jest": "^23.6.0",
```
rajivshah3: `jest` and `babel-jest` are bringing in vulnerable versions of `braces`, which is causing the Snyk test to fail. We need to force `braces@^2.3.1` to solve this issue. A package like https://www.npmjs.com/package/npm-force-resolutions might help, but that seems to work for lockfiles and not shrinkwraps. If I can figure out how to make it work for shrinkwraps I'll make a PR

rihardsgravis: You can achieve that by editing the shrinkwrap file manually and change the child dependency version there.

rajivshah3: Would it be overwritten by `npm install`?

rihardsgravis: It should not be overwritten - it's the opposite - `npm install` uses the shrinkwrap file to resolve the exact package versions (the only reason to have a shrinkwrap file at all).

rihardsgravis: So, jest released `24.0.0` two days ago. We are ok now 👍

rajivshah3:
@rihardsgravis I resolved the vulnerable dependency issue in https://github.com/facebook-atom/jest-electron-runner/commit/6f3e2979d7076171d7263517b34889036e384953. Just waiting for them to create a new release so the Snyk tests will pass

PR #1018, opened 2019-02-05
Realm data encryption

laumair:
# Description

Add encryption for realm database file . 

## Type of change
- Enhancement

# How Has This Been Tested?

- Manually tested iOS debug

# Checklist:


cvarley100 (approved):
LGTM!

src/mobile/src/libs/keychain.js
```diff
@@ -119,9 +121,43 @@ export const deleteTwoFactorAuthKeyFromKeychain = async () => {
...
+ * Sets realm encryption key.
+ *
+ * @method setRealmEncryptionKeyInKeychain
+ * @param {array} encryptionKey
+ *
+ * @returns {Promise}
+ */
+export const setRealmEncryptionKeyInKeychain = (encryptionKey) => {
+    return encodeBase64(encryptionKey).then((encodedEncryptionKey) =>
+        keychain.set(ALIAS_REALM, `Trinity-${getVersion()}`, encodedEncryptionKey),
```
rajivshah3, line 157: This key should probably be encrypted, because an application that spoofs the bundle ID could potentially read this value (this is only possible on a jailbroken device). I'm not sure what we would encrypt it with if we need it at load though 🤔

src/desktop/src/libs/realm.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Gets encryption key for realm.
+ * - Checks keychain if there is already an encryption key stored for realm data.
+ * - Returns encryption key stored in keychain (if exists).
+ * - Otherwise generates a new one and stores it in keychain.
+ *
+ * @method getEncryptionKey
+ * @returns {Promise}
+ */
+export const getEncryptionKey = () => {
```
laumair, line 14: @rihardsgravis Please let me know if you have different ideas about organisation of this function. This could also be placed with the other utils but I kept it separate to follow the mobile pattern.

src/mobile/src/libs/keychain.js
laumair: You are right. 

> I'm not sure what we would encrypt it with if we need it at load though
That's the problem. We don't have the password hash at this point. 

After having a discussion with @cvarley100 , we concluded that it's _okay_ to store it this way because we do warn users about the risks associated with using Trinity on a jailbroken/rooted device.

src/desktop/src/libs/realm.js
```diff
@@ -0,0 +1,27 @@
...
+    return Electron.readKeychain(ALIAS_REALM).then((encryptionKey) => {
+        if (encryptionKey === null) {
+            return Promise.resolve(randomBytes(64)).then((bytes) =>
+                Electron.setKeychain(ALIAS_REALM, base64.fromByteArray(encryptionKey)).then(
+                    () => new Uint8Array(bytes),
+                ),
+            );
+        }
+
+        return base64.toByteArray(encryptionKey);
```
rihardsgravis: There is no need to add an additional dependency for the string transformations. An `Uint8Array`  can be transformed to string via `encryptionKey.toString()` and back to Uint8Array as `Uint8Array.from(encryptionKey.split(","))`

rihardsgravis (changes requested):

rihardsgravis:
At https://github.com/laumair/trinity-wallet/blob/56e9b4202de688d98315f754fcabfd4ec8016440/src/desktop/src/libs/SeedStore/Keychain.js#L132 there should be also a check to skip `ALIAS_REALM`, otherwise password change would result in an error.

src/desktop/src/libs/realm.js
laumair: @rihardsgravis Addressed in https://github.com/iotaledger/trinity-wallet/pull/1018/commits/039a2fa4560fb963e1ae21101f62c615ad0e93bc

src/mobile/src/libs/keychain.js
rajivshah3: Sounds good 👍

rajivshah3 (approved):

laumair:
@rihardsgravis 

> At https://github.com/laumair/trinity-wallet/blob/56e9b4202de688d98315f754fcabfd4ec8016440/src/desktop/src/libs/SeedStore/Keychain.js#L132 there should be also a check to skip `ALIAS_REALM`, otherwise password change would result in an error.

Done in https://github.com/iotaledger/trinity-wallet/pull/1018/commits/1fbc64d071e42eb3bda51521f822d30c7f531153

laumair:
@rihardsgravis Ready for another review when you get a chance

rihardsgravis (approved):

PR #995, opened 2019-01-28
Mobile/iOS: Fix zoom calculation

rajivshah3:
# Description
Fixes zoom calculation on iOS, preventing cases where the QR scanner zooms in too much.

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

# How Has This Been Tested?

- Tested manually on device

# Checklist:

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] For changes to `mobile` that include native code (including React Native modules): I have verified that both iOS and Android successfully build in both `Debug` and `Release` modes

PR #375, opened 2018-09-15
Implement Realm Database

rajivshah3:
# Description

Closes #311 

## Type of change

- Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?

- Test A
- Test B

# Checklist:

- [ ] My code follows the style guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] For changes to `mobile` that include native code (including React Native modules): I have verified that both iOS and Android successfully build in both `Debug` and `Release` modes
- [ ] For changes to `shared`: If applicable, I have verified that my changes are implemented correctly in `desktop` and `mobile`

PR #1029, opened 2019-02-07
Mobile Release 0.4.1 (41)

cvarley100:
# Changelog 

- Add Realm database (#375)
- Encrypt Realm database (#1018)
- Fix iOS  zoom calculation (#995)
- Fix iPhone X (and variants) modal visual bug (#1018)
- Fix error log touch radius (#1018)
- Fix a bug where the full bundle is not stored on failed broadcast (#1018)
- Increase timeout for attachToTangle network requests (#989)
- New Crowdin translations (#981, #985, #1026)

### DIFF
--- a/src/mobile/src/ui/routes/entry.js
+++ b/src/mobile/src/ui/routes/entry.js
@@ -203,7 +203,7 @@ onAppStart()
             const { settings: { versions, completedMigration } } = reduxState;
 
             if (
-                versions.buildNumber < 41 &&
+                versions.buildNumber < 42 &&
                 completedMigration === false &&
                 // Also check if there is persisted data in AsyncStorage that needs to be migrated
                 // If this check is omitted, the condition will be satisfied on a fresh install.
--- a/src/mobile/android/app/build.gradle
+++ b/src/mobile/android/app/build.gradle
@@ -92,7 +92,7 @@ android {
     buildToolsVersion '27.0.3'
     defaultConfig {
         applicationId "com.iota.trinity"
-        versionCode 40
+        versionCode 41
         versionName "0.6.1"
         minSdkVersion 21
         targetSdkVersion 26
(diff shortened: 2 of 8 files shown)


src/mobile/src/ui/routes/entry.js
```diff
@@ -203,7 +203,7 @@ onAppStart()
             const { settings: { versions, completedMigration } } = reduxState;
 
             if (
-                versions.buildNumber < 41 &&
+                versions.buildNumber < 42 &&
```
rajivshah3, line 206: This should stay at 41, otherwise it may trigger another migration in a subsequent update

src/shared/libs/versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "desktopBlacklist": ["0.3.4", "0.3.5", "0.3.6", "0.4.0", "0.4.1"],
     "mobileBlacklist": ["20", "21", "22", "25", "26", "27"],
-    "latestMobile": "29",
+    "latestMobile": "41",
```
rajivshah3: Wouldn't this trigger an update notification for beta users who can't access this version?

rajivshah3 (changes requested):

src/shared/libs/versions.json
cvarley100: It's not being merged to develop, but good point. This shouldn't be updated until after release

src/mobile/src/ui/routes/entry.js
cvarley100: Na it should be `< 42` because that means Realm migration will trigger for this version

rajivshah3: Ooh my bad, I didn't realize that these held the current version and not the previous one

rajivshah3 (approved):

PR #1041, opened 2019-02-07
Mobile: Add retry button, error log and change node to Realm migration

cvarley100:
# Description

- Adds retry button, error log and change node to Realm migration

## Type of change

# How Has This Been Tested?

Tested on iOS simulator

# Checklist:

- [ ] My code follows the style guidelines for this project


src/mobile/src/ui/components/Migration.js
```diff
@@ -59,18 +73,37 @@ class Migration extends Component {
...
         this.props.startTrackingProgress(ProgressSteps.migration);
-
         this.props.migrate(reduxPersistStorageAdapter);
     }
 
     componentWillReceiveProps(newProps) {
         if (!this.props.completedMigration && newProps.completedMigration) {
             this.navigateToLoadingScreen();
         }
+        if (this.props.notificationLog !== newProps.notificationLog) {
```
laumair: `notificationLog` is an array. Perhaps we should be comparing size here (Also considering that there could be previous error logs).

src/mobile/src/ui/components/Migration.js
```diff
@@ -107,6 +140,25 @@ class Migration extends Component {
         });
     }
 
+    /**
+     * Changes to a random node
+     *
+     * @method changeNode
+     */
+    changeNode() {
+        this.props.setFullNode(...sample(this.props.nodes, 1));
```
laumair: [sample](https://lodash.com/docs/4.17.11#sample) takes a single parameter. Perhaps you meant to use [sampleSize](https://lodash.com/docs/4.17.11#sampleSize) here?

laumair (changes requested):

src/mobile/src/ui/components/Migration.js
rajivshah3: Would this trigger if the user cleared the error logs?

laumair (approved):

PR #1048, opened 2019-02-08
Mobile Release 0.6.1 (42)

cvarley100:
# Changelog

* Disable iOS pop gesture (#1041)
* Add retry button, notification log and node change to realm migration (#1041)
* Fix bug where unspent addresses are reported as as spent when reattaching (#1046)
* Fix account name change failure (#1045)


laumair:
LGTM!

PR #1090, opened 2019-02-17
Mobile: Update Argon2iOS

cvarley100:
# Description

- Update Argon2iOS to accept Int8 array

## Type of change

# How Has This Been Tested?

- Tested on iOS simulator

# Checklist:

### DIFF
--- a/src/mobile/ios/Argon2/Argon2IOS.m
+++ b/src/mobile/ios/Argon2/Argon2IOS.m
@@ -11,7 +11,7 @@
 
 @interface RCT_EXTERN_MODULE(Argon2IOS, NSObject)
 // Export hash method to RN
-RCT_EXTERN_METHOD(hash:(NSString *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
+RCT_EXTERN_METHOD(hash:(NSArray *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
 
 // Create a GCD queue for Argon2
 -(dispatch_queue_t)methodQueue
--- a/src/mobile/ios/Argon2/Argon2IOS.m
+++ b/src/mobile/ios/Argon2/Argon2IOS.m
@@ -11,7 +11,7 @@
 
 @interface RCT_EXTERN_MODULE(Argon2IOS, NSObject)
 // Export hash method to RN
-RCT_EXTERN_METHOD(hash:(NSArray *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
+RCT_EXTERN_METHOD(hash:(NSNumberArray *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
 
 // Create a GCD queue for Argon2
 -(dispatch_queue_t)methodQueue
--- a/src/mobile/ios/Argon2/Argon2IOS.swift
+++ b/src/mobile/ios/Argon2/Argon2IOS.swift
@@ -18,7 +18,7 @@ class Argon2IOS: NSObject {
   ///   - password: Password to hash
   ///   - resolve: A JS Promise resolve block
   ///   - reject: A JS Promise reject block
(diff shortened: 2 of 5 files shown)


src/mobile/ios/Argon2/Argon2IOS.swift
```diff
@@ -18,18 +18,28 @@ class Argon2IOS: NSObject {
...
   ///   - reject: A JS Promise reject block
-  @objc func hash(_ password: String, salt: String, params: [String: Any], resolver resolve: RCTPromiseResolveBlock, rejecter reject: RCTPromiseRejectBlock) {
+  @objc func hash(_ password: NSArray, salt: String, params: [String: Any], resolver resolve: RCTPromiseResolveBlock, rejecter reject: RCTPromiseRejectBlock) {
     // Validate parameters
+
     if params["t_cost"] is Int && params["m_cost"] is Int && params["parallelism"] is Int && params["hashLength"] is Int {
       // Resolve the hash
-      let h = Argon2Core.argon2Hash(password: password, salt: salt, params: params)
-      resolve(h)
+      let passwordLength = password.count as Int
```
rajivshah3, line 26: This cast isn't necessary
```suggestion
      let passwordLength = password.count
```

src/mobile/ios/Argon2/Argon2IOS.swift
```diff
@@ -18,18 +18,28 @@ class Argon2IOS: NSObject {
   ///   - password: Password to hash
   ///   - resolve: A JS Promise resolve block
   ///   - reject: A JS Promise reject block
-  @objc func hash(_ password: String, salt: String, params: [String: Any], resolver resolve: RCTPromiseResolveBlock, rejecter reject: RCTPromiseRejectBlock) {
+  @objc func hash(_ password: NSArray, salt: String, params: [String: Any], resolver resolve: RCTPromiseResolveBlock, rejecter reject: RCTPromiseRejectBlock) {
```
rajivshah3: We only want arrays of integers
```suggestion
  @objc func hash(_ password: [Int], salt: String, params: [String: Any], resolver resolve: RCTPromiseResolveBlock, rejecter reject: RCTPromiseRejectBlock) {
```

src/mobile/ios/Argon2/Argon2IOS.swift
```diff
@@ -18,18 +18,28 @@ class Argon2IOS: NSObject {
...
     // Validate parameters
+
     if params["t_cost"] is Int && params["m_cost"] is Int && params["parallelism"] is Int && params["hashLength"] is Int {
       // Resolve the hash
-      let h = Argon2Core.argon2Hash(password: password, salt: salt, params: params)
-      resolve(h)
+      let passwordLength = password.count as Int
+      var passwordInt8 = [Int8](repeating: 0, count: passwordLength)
+      for i in 0...passwordLength - 1 {
+        passwordInt8[i] = password[i] as! Int8
```
rajivshah3: Xcode gives a warning here: `Cast from 'UInt8' to unrelated type 'Int8' always fails`

Try replacing lines 27-30 with:

```swift
let passwordInt8 = password.map { Int8(bitPattern: $0) }
```

This will be a bit cleaner than a `for-in` loop and uses `init(bitPattern:)` instead of a force downcast (which basically tells the compiler "I know what I'm doing, let me try this anyway and if it doesn't work, trigger a crash")

src/mobile/ios/Argon2/Argon2IOS.m
```diff
@@ -11,7 +11,7 @@
 
 @interface RCT_EXTERN_MODULE(Argon2IOS, NSObject)
 // Export hash method to RN
-RCT_EXTERN_METHOD(hash:(NSString *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
+RCT_EXTERN_METHOD(hash:(NSArray *)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
```
rajivshah3: We only want arrays of integers
```suggestion
RCT_EXTERN_METHOD(hash:(NSArray<NSInteger*>*)password salt:(NSString *)salt params:(NSDictionary *)params resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject);
```

rajivshah3 (changes requested):
I wasn't able to test these changes since the build was failing

laumair:
Can't seem to build iOS.

cvarley100:
It doesn't build because it doesn't have the updated EntangledIOS methods

src/mobile/ios/Argon2/Argon2IOS.m
cvarley100: RCT Convert won't accept `NSArray<NSInteger*>*`, so I've changed it to `NSNumberArray`.

PR #1094, opened 2019-02-18
Desktop:  Add migration retry and node change functionality

rihardsgravis:
# Description

Add migration retry and node change functionality

## Type of change

# How Has This Been Tested?

Tested on macOS 10.13.6

# Checklist:

### DIFF
--- a/src/desktop/src/ui/global/Migration.js
+++ b/src/desktop/src/ui/global/Migration.js
@@ -6,7 +6,9 @@ import { connect } from 'react-redux';
 
 import { startTrackingProgress } from 'actions/progress';
 import { migrate } from 'actions/migrations';
+import { setFullNode } from 'actions/settings';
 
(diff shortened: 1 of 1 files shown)


src/desktop/src/ui/global/Migration.js
```diff
@@ -67,20 +88,51 @@ class Migration extends React.Component {
         ];
     }
 
+    changeNode = () => {
```
cvarley100, line 91: We could just do: `this.props.setFullNode(...sample(this.props.nodes));` with `lodash/sample` here

cvarley100 (approved):

src/desktop/src/ui/global/Migration.js
rihardsgravis: Desktop does not have `lodash` as dependency and additionally this method checks to not set the node that is already set.

PR #1167, opened 2019-03-05
Mobile/iOS: Rebuild Entangled framework

rajivshah3:
# Description

- Fix memory leaks
- Check for NULL before returning
- Switch to `memset_safe`
- Compile with `-O3` instead of `-Ofast`

## Type of change

# How Has This Been Tested?

- Tested on iOS simulator

# Checklist:


rajivshah3:
It's hard to tell whether signing and PoW became slower since we compiled with `-c opt` instead of `-Ofast` 🤔

cvarley100 (approved):

cvarley100 (approved):

PR #1172, opened 2019-03-07
Link up trit-based checksums

laumair:
# Description

Link up trit based (seed) checksums

## Type of change

- Enhancements

# How Has This Been Tested?

- Unit test coverage
- Manually tested iOS debug

# Checklist:


src/mobile/src/ui/components/Checksum.js
```diff
@@ -40,37 +42,49 @@ export class Checksum extends Component {
...
      * @return {string} Checksum or symbol to be shown
      */
-    getChecksumValue() {
-        const { seed } = this.props;
-        let checksumValue = '...';
+    static getValue(input) {
+        const sizeOfInput = size(input);
 
-        if (seed.length !== 0 && !seed.match(VALID_SEED_REGEX)) {
-            checksumValue = '!';
```
rajivshah3, line 50: Shouldn't we immediately alert the user that they've typed an invalid character? Instead they will have to type the entire seed, get an alert, and then look for the invalid character

laumair: @rajivshah3 I don't think Text Input for seed entry would allow any invalid characters. See https://github.com/iotaledger/trinity-wallet/blob/develop/src/mobile/src/ui/components/CustomTextInput.js#L215

rajivshah3 (approved):

PR #1174, opened 2019-03-10
Mobile: Fix "Realm is not a constructor" in Chrome debugger

rajivshah3:
# Description

When running in the Chrome debugger, `Realm` is undefined when `initialise` is called. This is because (for some reason) the assignment is never evaluated:
https://github.com/iotaledger/trinity-wallet/blob/29be9a12db3742188e5269916177bac791335d08/src/shared/storage/index.js#L58

When the Chrome debugger is detected, `Realm` (a local variable inside the `initialise` block) is assigned to the value of `getRealm`. This forces the debugger to call `getRealm`.

## Type of change

# How Has This Been Tested?

- Tested on iOS (`Debug` without debugger, `Debug` with debugger, and `Release`) 

# Checklist:

### DIFF
--- a/src/shared/storage/index.js
+++ b/src/shared/storage/index.js
@@ -37,6 +37,9 @@ const getStoragePath = (schemaVersion = SCHEMA_VERSION) =>
 // Initialise realm instance
 let realm = {}; // eslint-disable-line import/no-mutable-exports
 
+// Initialise Realm constructor as null and reinitialise after importing the correct (platform) Realm dependency
+let Realm = null;
+
 /**
  * Imports Realm dependency
  *
@@ -54,8 +57,6 @@ export const getRealm = () => {
 
     return Electron.getRealm();
 };
-
-const Realm = getRealm();
 
 /**
  * Model for Account.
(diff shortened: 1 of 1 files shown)


src/shared/storage/index.js
```diff
@@ -775,7 +775,13 @@ const purge = () =>
  */
 const initialise = (getEncryptionKeyPromise) =>
     getEncryptionKeyPromise().then((encryptionKey) => {
-        realm = new Realm(assign({}, config, { encryptionKey }));
+        // For some reason `getRealm` doesn't get called before this function when running in the Chrome debugger
+        if (__MOBILE__ && /Chrome/.test(navigator.userAgent)) {
+            const Realm = getRealm();
```
laumair: `initialise` is called on app entry (both desktop & mobile). When Chrome debugging is enabled, `initialise` is called before `Realm` constructor is assigned to the global variable. 

This solution removes the missing constructor error but throws an exception [here](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/storage/index.js#L797) because the global constructor remains `undefined`.

laumair: I suggest to remove the _direct_ global assignment and assign correct constructor to `Realm` inside the initialisation function. This should also work on desktop because of [this](https://github.com/iotaledger/trinity-wallet/blob/develop/src/desktop/src/index.js#L43).

laumair (changes requested):

src/shared/storage/index.js
laumair: @rajivshah3 

I've made the above [changes](https://github.com/iotaledger/trinity-wallet/commit/78dc7fe94f3984066d43ec0e79fc6e77af5b267f). Lemme know your thoughts on the approach and I'll open a PR against this branch.

rajivshah3: Looks good! I like that approach better

laumair: @rajivshah3 https://github.com/iotaledger/trinity-wallet/pull/1177

laumair (approved):

PR #1175, opened 2019-03-11
Mobile/iOS: Argon2 hashing updates

rajivshah3:
# Description

- Bump CatCrypto to 0.3.0
- Return hash as `[UInt8]`
- Accept password as `[UInt8]` (converts to `String` on native side)

## Type of change

# How Has This Been Tested?

- Tested hashing during onboarding on iOS
- Tested hashing during login (with onboarding completed on `develop` branch) on iOS

# Checklist:

### DIFF
--- a/src/mobile/ios/Argon2/Argon2Core.swift
+++ b/src/mobile/ios/Argon2/Argon2Core.swift
@@ -39,14 +39,14 @@ struct Argon2Core {
   ///   - salt: Salt to be used
   ///   - params: Parameters to initialize Argon2 with
   /// - Returns: Hashed password
-  static func argon2Hash(password: String, salt: String, params: [String: Any]) -> String {
+  static func argon2Hash(password: String, salt: String, params: [String: Any]) -> [UInt8] {
     // Initialize Argon2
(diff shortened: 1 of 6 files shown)


mgeier63:
Re "[Pass UInt8 array to Argon2 instead of String](https://github.com/iotaledger/trinity-wallet/pull/1175/files#diff-71a3e664432a5bc427d39baff98d3538)" / commit.

I'm afraid that's not what the code does. The [lodash] `values(...)` call converts the UInt8Array to a regular JS array before passing it on to the native hash function - and that array is never cleaned up.
Likewise, the returned value is also a regular JS array and that's why you're converting it with `new Uint8Array(result)` already.

Unfortunately, there doesn't seem to be a way to directly pass UInt8Arrays (or other typed arrays for that matter) through the RN native bridge. Without patching / extending that bridge, you're limited to either a String or a regular JS array, both of which having "issues" as we know.

rajivshah3:
@mgeier63 that's a good point. I'm looking into whether we can allow `Uint8Array`s to be created and managed via the JavaScriptCore API: https://developer.apple.com/documentation/javascriptcore/1644597-jsobjectmaketypedarray

laumair (approved):
Looks good.

PR #1185, opened 2019-03-18
Desktop: Fix snapshot transition

### LINKED ISSUE #1134: snapshot transition always shows 0i
### Bug description
The snapshot transition always shows 0i as detected balance although my seed doesn't have many used addresses and has an address with iotas on it. I clicked about 20 times on `No` as it showed me 0i as balance. The same happened on another seed too.

### Environment Specs

- OS version: macOS 10.14.3

- Trinity version: 0.5.0-RC1

### Steps To Reproduce
1. start snapshot transition
2. click on No as it shows 0i as balance

### Expected behaviour
should show the correct balance after clicking `No` for a few times.

### Actual behaviour
always shows 0i

rajivshah3:
# Description
Passes addresses from selector instead of from `addressData` object

Fixes #1134 

## Type of change

# How Has This Been Tested?

- Tested on macOS (debug)

# Checklist:

### DIFF
--- a/src/desktop/src/ui/views/settings/account/Tools.js
+++ b/src/desktop/src/ui/views/settings/account/Tools.js
@@ -13,6 +13,8 @@ import {
     generateAddressesAndGetBalance,
 } from 'actions/wallet';
 
+import { getAddressesForSelectedAccount } from 'selectors/accounts';
+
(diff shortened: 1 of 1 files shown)


laumair (approved):
👍

PR #1224, opened 2019-03-29
Mobile/iOS: Fix NSLocationAlwaysUsageDescription purpose string

rajivshah3:
# Description

Fixes `NSLocationAlwaysUsageDescription` purpose string to satisfy App Store requirements. Build 50 was rejected because this was not specified in the Info.plist.

**Note**: This string should not appear during normal app use because Trinity never needs access to your location. This is simply to satisfy App Store requirements.

## Type of change

# How Has This Been Tested?
N/A

# Checklist:


rajivshah3:
Marking as critical because iOS builds will be rejected without this fix

cvarley100 (approved):

PR #1239, opened 2019-04-01
Mobile: Ensure seed reentry input only accepts A-Z9

cvarley100:
# Description

- Ensure seed reentry input only accepts A-Z9

## Type of change

# How Has This Been Tested?

Tested on iOS Simulator

# Checklist:

### DIFF
--- a/src/mobile/src/ui/views/onboarding/SeedReentry.js
+++ b/src/mobile/src/ui/views/onboarding/SeedReentry.js
@@ -87,7 +87,7 @@ class SeedReentry extends Component {
     constructor() {
         super();
         this.state = {
-            reenteredSeed: null,
+            reenteredSeed: '',
         };
     }
 


src/mobile/src/ui/views/onboarding/SeedReentry.js
```diff
@@ -87,7 +87,7 @@ class SeedReentry extends Component {
     constructor() {
         super();
         this.state = {
-            reenteredSeed: null,
+            reenteredSeed: '',
```
rajivshah3, line 90: I think this change might break QR code scanning. On line 158 we set `reenteredSeed` to a trit value:

```js
this.setState({ reenteredSeed: trytesToTrits(data) });
```
Since `trytesToTrits` returns an `Int8Array` I don't think we can initialize it to an empty string

cvarley100: It doesn't break it

rajivshah3 (approved):

PR #1244, opened 2019-04-02
Mobile: Minor changes to seed text input

cvarley100:
# Description

- Unmask seed only on long press
- Allow and autocapitalise lower case chars on seed input

## Type of change

# How Has This Been Tested?

- Tested on Android and iOS devices

# Checklist:


rajivshah3:
Maybe we should add some sort of note to the user (e.g. `press and hold to reveal`)?

rajivshah3 (approved):

PR #1291, opened 2019-04-11
Mobile: Fix seed text input issues

cvarley100:
# Description

- Remove push to mask seed text input  - Push to mask clears the iOS text input and is not needed on Android due to FlagSecure
- Fix Android letter duplication  - `text.toUpperCase()` causes letter duplication and `autoCapitalise='letters'` does not work with `secureTextEntry`

## Type of change

# How Has This Been Tested?

Tested on Android and iOS simulators

# Checklist:


laumair (approved):
Looks good! Thanks.

src/mobile/src/ui/components/CustomTextInput.js
```diff
@@ -537,6 +537,7 @@ class CustomTextInput extends Component {
                         onChangeText={(text) => this.onChangeText(text)}
                         selectionColor={theme.input.alt}
                         underlineColorAndroid="transparent"
+                        autoCorrect={false}
```
rajivshah3: Won't this affect all text fields, even ones where users would expect auto-correct (e.g. `message` field when sending a transaction)?

cvarley100: Good catch. This was a test and wasn’t meant to be committed.

PR #1341, opened 2019-04-17
Allow (alpha) users to export wallet state

laumair:
# Description

This PR will allow alpha stage users to export state. 

## Type of change

# How Has This Been Tested?

- Manually tested iOS (debug)
- 

# Checklist:

### DIFF
--- a/src/mobile/src/ui/views/wallet/AdvancedSettings.js
+++ b/src/mobile/src/ui/views/wallet/AdvancedSettings.js
@@ -139,6 +139,11 @@ export class AdvancedSettings extends PureComponent {
                 function: () => this.props.setSetting('snapshotTransition'),
             },
             { name: t('manualSync'), icon: 'sync', function: () => this.props.setSetting('manualSync') },
+            {
+                name: t('exportState'),
+                icon: 'copy',
+                function: () => this.props.setSetting('stateExport'),
+            },
             { name: 'separator' },
             { name: t('settings:reset'), icon: 'trash', function: this.reset },
             { name: 'back', function: () => this.props.setSetting('mainSettings') },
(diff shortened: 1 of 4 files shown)


src/mobile/src/ui/views/wallet/AdvancedSettings.js
```diff
@@ -165,6 +170,6 @@ const mapDispatchToProps = {
     generateAlert,
 };
 
-export default withNamespaces(['advancedSettings', 'settings', 'global'])(
+export default withNamespaces(['advancedSettings', 'settings', 'global', 'stateExport'])(
```
rajivshah3, line 173: I'm not sure if this needs to be added since the `exportState` string is in the `global` namespace, which is already specified

src/mobile/src/ui/views/wallet/StateExport.js
```diff
@@ -0,0 +1,198 @@
...
+                                    accounts: pick(accounts, ['onboardingComplete', 'accountInfo']),
+                                    settings: pick(settings, ['versions']),
+                                    notificationLog,
+                                    storageFiles: filter(flatMap(files), (name) => includes(name, 'realm')),
+                                },
+                                null,
+                                4,
+                            ),
+                        );
+                        this.props.generateAlert('success', t('exportSuccess'), t('exportSuccessExplanation'));
```
rajivshah3, line 144: We should trigger the share menu after writing the file, especially since iOS users do not generally have access to the filesystem (so they will not be able to access the file we just created). We can use a similar approach to the one we use for exporting a SeedVault:

https://github.com/iotaledger/trinity-wallet/blob/e3290d2333eb46d986336b90259168cda210cfe2/src/mobile/src/ui/components/SeedVaultExportComponent.js#L296-L307

PR #1368, opened 2019-04-21
Make sure we reorder account indexes (in realm) on account deletion

laumair:
# Description

Deleting account doesn't reorder indexes in realm db leading to gaps in account indexes. This PR fixes the issue. 

## Type of change

# How Has This Been Tested?

- Manually tested iOS (debug)

# Checklist:

### DIFF
--- a/src/shared/storage/index.js
+++ b/src/shared/storage/index.js
@@ -1,13 +1,16 @@
 /* global Electron */
 import assign from 'lodash/assign';
 import each from 'lodash/each';
+import find from 'lodash/find';
 import includes from 'lodash/includes';
 import isArray from 'lodash/isArray';
 import isEmpty from 'lodash/isEmpty';
 import isUndefined from 'lodash/isUndefined';
 import map from 'lodash/map';
 import merge from 'lodash/merge';
+import orderBy from 'lodash/orderBy';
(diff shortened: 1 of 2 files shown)


src/shared/storage/index.js
```diff
@@ -148,7 +170,27 @@ class Account {
      * @param {string} name
      */
     static delete(name) {
-        realm.write(() => realm.delete(Account.getObjectForId(name)));
+        realm.write(() => {
+            const accountsBeforeDeletion = Account.getDataAsArray();
+            const accountForDeletion = find(accountsBeforeDeletion, { name });
```
rajivshah3, line 175: Would it be easier to simply do this?

```js
const accountForDeletion = Account.getObjectForId(name);
```

laumair: [getDataAsArray](https://github.com/iotaledger/trinity-wallet/blob/7da5732886cea2d3640ded8d81bf50449b027188/src/shared/storage/index.js#L73) already gets all the accounts in realm db. I think it's better to just find the account (to delete) rather than making another query to realm db.

rajivshah3: Ah, I see

rajivshah3 (approved):

PR #1406, opened 2019-04-26
Shared: Fix node sync check

rajivshah3:
# Description
This fixes the node sync check so that it allows a range of milestones. If the `latestMilestoneIndex` is `5` and the `latestSolidSubtangleMilestoneIndex` is `4`, the node health check should now return `true` because the `latestSolidSubtangleMilestoneIndex` is between `3` (`latestMilestoneIndex - MAX_MILESTONE_FALLBEHIND`) and `5`.

## Type of change

# How Has This Been Tested?

- Updated unit tests

# Checklist:


laumair (approved):
Good catch! 💯

PR #1418, opened 2019-04-27
Fix RC1 -> RC2 installation issue on Windows

laumair:
Installing RC2 on top of RC1 build leads to an app crash throwing `Error: resource busy or locked`. This commit fixes the issue.

[Realm.schemaVersion(path)](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/storage/index.js#L817) creates unnecessary realm files. Therefore we had to manually [remove](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/storage/index.js#L852-L859) these files. On Windows 10 however, it was conflicting with the `Realm.deleteFile` operation [here](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/storage/index.js#L797) which was rather unnecessary as we are manually deleting the files anyways in the next step.
# Description

## Type of change

- Bug fix 

# How Has This Been Tested?

- Manually tested Windows 10

# Checklist:

### DIFF
--- a/src/shared/storage/index.js
+++ b/src/shared/storage/index.js
@@ -793,8 +793,6 @@ const migrateToNewStoragePath = (config) => {
     });
 
     oldRealm.write(() => oldRealm.deleteAll());
-
-    Realm.deleteFile(config);
 };
 
 /**


rajivshah3 (approved):
Looks good! Tested on macOS

PR #1487, opened 2019-05-08
Mobile: Add end-to-end tests

rajivshah3:
# Description
Adds end-to-end tests for mobile

## Type of change

# How Has This Been Tested?

- Ran E2E tests

# Checklist:


christian-bromann:
I am probably late to the game but I recommend to use [WebdriverIO](https://webdriver.io/) for the e2e test suite for the following reasons:

- It allows to write a single test that can be executed for Desktop (Electron Apps) and Mobile (Android/iOS) environments using [Appium](https://appiumconf.com/)
- [Page objects](https://webdriver.io/docs/pageobjects.html) can be written so that they fetch the element according to the environment the test is running
- Automatically runs test in parallel
- Allows to run tests in an cloud environment like Sauce Labs (as an employee I am happy to provide you with a free open source account that will fit your needs)

I will have to mention that I am the maintainer of the WebdriverIO project and a fan of IOTA 😉

rajivshah3:
Hi @christian-bromann , that sounds great! Does Appium support React Native apps? This was the primary reason we chose to use [Detox](https://github.com/wix/Detox) since we can write tests in JS and have them execute on both Android and iOS.

christian-bromann:
> Does Appium support React Native apps?

Yes it does. The problem I see with Detox is that it is an own automation implementation that only works in mobile (Android not even fully supported). Appium is based on a the WebDriver standard that allows you to write tests for all end devices (desktop and mobile). In addition to that you don't need to modify your application to be able to test it. Appium uses native automation frameworks that allow you to automate everything on the phone.

PR #1549, opened 2019-05-17
Desktop: Catch and display preload and initialisation errors

rihardsgravis:
# Description

Show to the user:
 - Error output if the wallet could not be initialised correctly
 - Errors on the Electron preload script 

Previously the wallet did not display the main window, but the process was running in the background.

## Type of change

# How Has This Been Tested?

Tested on macOS

# Checklist:

### DIFF
--- a/src/desktop/src/index.js
+++ b/src/desktop/src/index.js
@@ -25,84 +25,92 @@ import FatalError from 'ui/global/FatalError';
 
 import settings from '../package.json';
 
-export const bugsnagClient = bugsnag({
-    apiKey: 'fakeAPIkey',
-    appVersion: settings.version,
-    interactionBreadcrumbsEnabled: false,
-    collectUserIp: false,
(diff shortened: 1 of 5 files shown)


src/desktop/src/index.js
```diff
@@ -21,84 +21,96 @@ import Index from 'ui/Index';
...
+
+                // Start Tray application if enabled in settings
+                const isTrayEnabled = get(data, 'settings.isTrayEnabled');
+                Electron.setTray(isTrayEnabled);
+
+                // Show Wallet window after inital store update
+                Electron.focus();
+            })
+
+            .catch((err) => {
```
rajivshah3, line 89: Are we able to send the error to Bugsnag as well?

rihardsgravis: Added

rajivshah3 (approved):
Looks good 👍

PR #1605, opened 2019-05-24
Shared: Fix `getSelectedAccountMeta` selector

rihardsgravis:
# Description

Fix `getSelectedAccountMeta` to not raise fatal error if `typeof account === "undefined"`

Happens when first onboarding is finished and initial load raises an error. The desktop wallet returns to Login screen, where this selector is used in props, but no account is added yet.

## Type of change

# How Has This Been Tested?

Tested on macOS

# Checklist:

### DIFF
--- a/src/shared/selectors/accounts.js
+++ b/src/shared/selectors/accounts.js
@@ -99,7 +99,7 @@ export const selectLatestAddressFromAccountFactory = (withChecksum = true) =>
  **/
 export const getSelectedAccountMeta = createSelector(
     selectAccountInfo,
-    (account) => account.meta || { type: 'keychain' },
+    (account) => get(account, 'meta') || { type: 'keychain' },
 );
 
 /**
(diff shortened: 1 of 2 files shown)


src/shared/selectors/accounts.js
```diff
@@ -99,7 +99,7 @@ export const selectLatestAddressFromAccountFactory = (withChecksum = true) =>
  **/
 export const getSelectedAccountMeta = createSelector(
     selectAccountInfo,
-    (account) => account.meta || { type: 'keychain' },
+    (account) => get(account, 'meta'),
```
rajivshah3, line 102: If you want to preserve the existing functionality where it will return `'keychain'` as a fallback:

```suggestion
    (account) => get(account, 'meta', { type: 'keychain' }),
```

rihardsgravis: Returning `undefined` would be a more correct behaviour.

laumair (approved):
Looks good to me!

PR #1609, opened 2019-05-25
Shared: Use new realm updateMode

rajivshah3:
# Description
Realm introduced new update modes when passing a third parameter to `realm.create`. The boolean parameters are now deprecated (`true` is `'all'` and `false` is `'never'`). The new `'modified'` option may help reduce unnecessary writes or duplicate objects.

> - 'never': Objects are only created. If an existing object exists, an exception is thrown. This is the default value.
> - 'all': If an existing object is found, all properties provided will be updated, any other properties will remain unchanged.
> - 'modified': If an existing object exists, only properties where the value has actually changed will be updated. This improves notifications and server side performance but also have implications for how changes across devices are merged. For most use cases, the behaviour will match the intuitive behaviour of how changes should be merged, but if updating an entire object is considered an atomic operation, this mode should not be used.

See docs: https://realm.io/docs/javascript/2.27.0/api/Realm.html#create

## Type of change

# How Has This Been Tested?

- Tested on macOS (debug)
- Refreshed history and performed manual sync to verify that duplicate objects were not created (on `develop`, duplicate objects will be created when refreshing history or performing a manual sync)

# Checklist:


laumair (approved):
Also tested on iOS simulator iPhone 7 Plus

Issue #1639, opened 2019-05-28
Message "New update available" doesn't close automatically after update process was started


ratiodep:
### Description
Close the red message "New update (...)" after the user started the update procedure

Trinity 0.71
Win 10 64 bit

rihardsgravis:
Closed in #1705

Issue #1626, opened 2019-05-27
Error message when checking for updates without internet connection


ratiodep:
This issue has already been submitted with #282 but happens now again (maybe never fixed in Win 7?)

![bug_nointernet](https://user-images.githubusercontent.com/50994219/58399622-0b355500-8059-11e9-885b-def3a6ec9b16.png)

### Environment Specs
What device are you running Trinity on?

- Manufacturer: Dell
- Model: Precision 7720
- OS version: Win 7 64 bit

- Trinity version: 0.51

### Steps To Reproduce
1. Cut internet connection
2. Start Trinity 
3. Start manual update

rihardsgravis:
Closed in #1705

Issue #1625, opened 2019-05-26
 Fatal error launching Trinity desktop wallet v0.5.1. Linux/Ubuntu


Nick2567:
### Bug description
I was prompted to update Trinity from v0.4.6. The update completed and a message said "Trinity will now restart" but the message did not disappear so I restarted Trinity manually. Upon launching Trinity-desktop-0.5.1. a fatal error message displays. I still have access to Trinity, Edit, Help menus.

### Environment Specs
What device are you running Trinity on?

- Manufacturer: Dell
- Model: Inspiron
- OS version: Linux Ubuntu 16.04

What build of Trinity are you running? 

- Trinity version: Trinity-desktop-0.5.1.AppImage

### Steps To Reproduce
1. Double click on Trinity-desktop-0.5.1.AppImage
2. Results in error message indicated below.

### Expected behaviour
Wallet should launch normally as in v0.4.6.

### Actual behaviour
Launching wallet results in fatal error as described below.

### Errors
Error launching wallet.
There was a fatal error launching the wallet.
Failed to load Electron preload script
{"error":"Uncaught ReferenceError: setImmediate is not defined"}

rajivshah3:
Was there a stacktrace in the error message, or did it only say `Uncaught ReferenceError: setImmediate is not defined`?

Nick2567:
The first time I launched 0.5.1 there was a more detailed error message but now I just see the image below. I have access to the menu for Trinity --> Error Log but nothing appears.

![image](https://user-images.githubusercontent.com/32357455/58430429-452a4980-80a1-11e9-85cb-159ad0a2e8e0.png)

rajivshah3:
Hmm. Do you remember anything from the detailed message?

Nick2567:
No, unfortunately not. After restarting the wallet again I though it would either run without error or the same error message would appear.

Do you think I should download v0.4.6 again or download v0.5.1 and restore from seed vault?

Nick2567:
Launched Trinity 0.5.1. (error message above still displaying)
Went to Trinity --> Check for Updates. A new Trinity version is available. Clicked Yes to update.
![image](https://user-images.githubusercontent.com/32357455/58752092-d08c4c00-84a0-11e9-987f-924eb910aa42.png)
Update completed but Trinity did not restart. Message stayed on screen.
![image](https://user-images.githubusercontent.com/32357455/58752102-fa457300-84a0-11e9-966b-4e3a746203c7.png)
[...]

Issue #1594, opened 2019-05-23
Trinity does not work correctly on Windows when username has special characters


rajivshah3:
### Bug description
We're seeing error reports where Realm initialization is failing with an error similar to one of the following:  
```
Unable to open a realm at path 'C:\Users\username\AppData\Roaming\Trinity\trinity.realm.management': make_dir() failed: No such file or directory Path:.
```
```
Unable to open a realm at path 'C:\Users\username\AppData\Roaming\Trinity\trinity.realm.management': make_dir() failed: Illegal byte sequence Path:.
``` 

This may be related to a bug in Realm where paths with special characters are not supported (https://github.com/realm/realm-js/issues/2319)

### Environment Specs
What device are you running Trinity on?

- Manufacturer: N/A
- Model: N/A
- OS version: Windows

What build of Trinity are you running? Please provide the full version (you can find it in Settings > About Trinity).

- Trinity version: v0.5.0 and later

### Steps To Reproduce
1. Create a user account with special characters in the username (Chinese, Korean, Cyrillic, and accented Latin characters are known to cause this issue)
2. Attempt to open the latest version of Trinity

### Expected behaviour
The app should open and function correctly

### Actual behaviour
The app might not function correctly

rajivshah3:
A possible workaround may be to create another user account (without special characters in the username) for using Trinity

rihardsgravis:
We should try to store Realm database in another, not user related path (e.g. the application path) for the Realm db.

tolotos-rgu:
I have the same behavior. My Win10 username is "first name""blank""last name" the first name conains a German Umlaut (ü)

suefkem:
Same issue here. Win10 Pro (1803), had the issue with trinity wallet after the last update. Username: Firstname(Whitespace)Lastname(with german Umlaut)

-  uninstalled it and deleted files located in appdata/roaming, downloaded and tried to install fresh -> same error. 
app from testflight/ios does somehow always reset and forces me to completly set up everything from scratch after I colse the app.

rajivshah3:
> app from testflight/ios does somehow always reset and forces me to completly set up everything from scratch after I colse the app.

That's odd 🤔 would you mind creating another issue for that?

suefkem:
> > app from testflight/ios does somehow always reset and forces me to completly set up everything from scratch after I colse the app.
> 
> That's odd 🤔 would you mind creating another issue for that?

[...]

PR #1671, opened 2019-05-31
Desktop 0.6.0

### LINKED ISSUE #523: Optional setting of "basic authentication" credentials for custom nodes
### Description
Integration of "basic authentication" as an option for added custom nodes.

### Motivation
Some users prefer to secure their nodes from the public via setting credentials to the IRI API port.
Most used type is the "basic authentication" standard of the HTTP protocol.

### Requirements
On the node section of the settings menu, there could simply be two optional fields "user" and "password" under the "custom node" field. If both contain content, then include the credentials on each http request to the IRI API port of this node.

### Open Questions (optional)
Is it intended to include this feature, are there any drawbacks?

Maybe

rihardsgravis:
# Description

- New: Automatic node management and lots of configurability for advanced users (#1509)
- Update: Auto updates functionality improvements (#1705)
- Update: Add several wallet initialisation error explanations (#1727)
- Update: Enable Hardened Runtime on MacOS (#1751)
- Fix: Ignore system proxy setting is not saved (#1689)
- Fix: Wallet language resets to English (#1704)
- Fix: Windows installation should remove appData on uninstall (#1726)
- Fix: Allow changing install directory (#1726)
- Fix: Wallet does not work on Windows when username has special characters (#1767)
- Update: Crowdin translations (#1657, #1680, #1719)

Closes #523, #544, #722, #892, #894, #1594,  #1625, #1626, #1639, #1683, #1685, #1692

## Type of change

- New release

# Checklist:


src/desktop/__tests__/ui/global/__snapshots__/about.test.jsx.snap
```diff
@@ -18,14 +18,34 @@ exports[`About component Render the component 1`] = `
...
+      0.5.2
        
       <small>
         BETA
       </small>
     </h2>
     <article>
       <Scrollbar>
+        <h5>
+          0.5.2
```
rajivshah3: We might want to think about how we can make this test less flaky -- it will fail and require a snapshot update each time we edit the changelog

rihardsgravis: We should move the change log to `CHANGELOG.txt` and load the content from there. Markdown is already supported, so no big changes required. I'll do a PR.

cvarley100 (approved):

PR #1726, opened 2019-06-07
Desktop: Windows installer updates

### LINKED ISSUE #722: Change install directory during the installation process (or better: make Trinity portable)
On a laptop I want to protect my data as good as possible, using a protection method of MY choice. That is why I have all files and sensitive programs (like home banking applications as well as crypto wallets) on a crypted volume, which I mount manual if needed after the OS has started.

Most of Electrum wallets for example having a portable version, which is super nice and easy to put on the save/secret volume. For Trinity I had to change all the Registry values in Windows pointing to a new folder on that volume. Yes, that works but it sucks to do this manually.

**PLEASE** give an option ("Expert options" or whatever) like ANY other installation routine does, so a user can choose its install directory during installation. Or to leave even less traces in the system, make Trinity portable! I really hate it in Windows if programs just install themself on C:\ without asking! Maybe most users don't care, but I will command my computer where he should put things. Please understand that, thanks :-)

rihardsgravis:
# Description

- Delete `appData` on uninstall

## Type of change

# How Has This Been Tested?

Tested on Windows 10

# Checklist:

### DIFF
--- a/src/desktop/package.json
+++ b/src/desktop/package.json
@@ -115,6 +115,11 @@
         }
       ]
     },
+    "nsis": {
+      "oneClick": false, 
+      "allowToChangeInstallationDirectory": true,
+      "deleteAppDataOnUninstall": true
+    },
     "win": {
       "publisherName": "IOTA Stiftung",
       "target": "nsis"
--- a/src/desktop/package.json
+++ b/src/desktop/package.json
@@ -116,8 +116,6 @@
       ]
     },
     "nsis": {
-      "oneClick": false, 
-      "allowToChangeInstallationDirectory": true,
       "deleteAppDataOnUninstall": true
     },
     "win": {


rajivshah3 (approved):

src/desktop/package.json
```diff
@@ -115,6 +115,11 @@
         }
       ]
     },
+    "nsis": {
+      "oneClick": false, 
+      "allowToChangeInstallationDirectory": true,
+      "deleteAppDataOnUninstall": true
```
rajivshah3, line 119: According to the [`electron-builder` docs](https://www.electron.build/configuration/nsis), `deleteAppDataOnUninstall` is only available when using the one-click installer

rihardsgravis: :( I'll leave `deleteAppDataOnUninstall` only then as it's more required than changing the installation directory.

PR #1811, opened 2019-06-18
Shared: Add fallback node list api endpoints

rihardsgravis:
# Description

Add fallback node list api endpoints and update `fetchRemoteNodes` functionality

## Type of change

# How Has This Been Tested?

Tested on macOS

# Checklist:


src/shared/libs/iota/utils.js
```diff
@@ -425,28 +425,44 @@ export const withRetriesOnDifferentNodes = (nodes, failureCallbacks) => {
...
-                return response.filter((node) => typeof node.node === 'string' && node.node.indexOf('https://') === 0);
+    };
+
+    let remoteNodes = [];
+
+    for (let index = 0; index < NODELIST_ENDPOINTS.length; index++) {
+        try {
+            const endPoint = NODELIST_ENDPOINTS[index];
+
+            const response = await fetch(endPoint, requestOptions, GET_NODELIST_REQUEST_TIMEOUT);
```
rajivshah3: `fetch` does not appear to support timeouts according to the [spec](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch). To achieve similar functionality we use `axios` on mobile:

https://github.com/iotaledger/trinity-wallet/blob/f3fab95c9f20a9afc21a36ecf8473ca664ae49b7/src/mobile/src/ui/routes/entry.js#L187-L208

cvarley100 (approved):

PR #1835, opened 2019-06-20
Shared: Fix bugs associated with quorum

cvarley100:
# Description

- Ensures quorum is not performed when explicitly turned off during send
- Correctly sync account if there is an error after broadcast
- Fixes associated incorrect `alreadySpentFrom` error message

Fixes #1538
Fixes #1386

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

# How Has This Been Tested?

Tested on iOS

# Checklist:

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] For changes to `mobile` that include native code (including React Native modules): I have verified that both iOS and Android successfully build in both `Debug` and `Release` modes
- [ ] For changes to `shared`: If applicable, I have verified that my changes are implemented correctly in `desktop` and `mobile`

Issue #1815, opened 2019-06-18
Desktop wallet launches with blank screen


DevTore:
### Bug description
No graphic on the screen after launching app

### Hardware Spec
MacBook Pro

### Expected behaviour
1. White Trinity/ IOTA logo appears and disappears on a dark blue background
2. login window etc is displayed

### Actual behaviour
1. White Trinity/ IOTA logo appears and disappears on a dark blue background
2. Display continues to show dark blue background only (no login window, etc.)

### Errors

![image](https://user-images.githubusercontent.com/42974518/59706199-f8ccb800-91cd-11e9-9eed-61a539275ffa.png)

<img width="1436" alt="Screen Shot 2019-06-18 at 1 34 47 PM" src="https://user-images.githubusercontent.com/42974518/59706336-46492500-91ce-11e9-852d-8e1ae1a95170.png">

t00k:
If you updated from a really old version, try downloading 0.4.6 (find it on releases for this github) and run it, then update from 0.4.6 to the latest version. Make sure to have your seeds backed up.

laumair:
@DevTore Did you update the wallet from a previous version?

DevTore:
I downloaded the latest version and replaced the old version. After opening the new version, there was just a dark blue screen after the iota symbol had disappeared. I tried different things, such as
- clicking the “check for newest version” in the drop down menu, but received the message that I already have the latest version
- Downloading the newest version (from the IOTA Trinity download page as well as directly from Github) and replacing, with the same result
- restarting computer and repeating steps above, with the same result

Since the Trinity app only shows dark blue, there is nowhere to enter the seed.

laumair:
@DevTore Do you (by any chance) remember the older version number of the application you were using?

DevTore:
Unfortunately, I do not know the older version number of the application.
However, I followed @t00k advise to download and install v04.6. When I ran it, my MIOTA balance also displayed correctly.
Subsequently, I used the scroll down menu "Look for updates" and let the application automatically update to the latest version, which worked.
BUT, when entering my password, a 'Data migration' screen appeared, with the message:
"Something went wrong - There was a problem migrating your data. Please try again"
AND the following message: 
"Due to performance and storage limitations, Trinity is using a new database for storing settings and account data. Please wait while Trinity automatically migrates your existing data."
Two option buttons are displayed underneath: 'Change Node' and 'Retry'
[...]

Issue #1422, opened 2019-04-28
Cannot read property 'meta' of undefined


rajivshah3:
## Error in Trinity Desktop

**TypeError** in **/index.html**
Cannot read property 'meta' of undefined

[View on Bugsnag](https://app.bugsnag.com/iota-foundation/trinity-desktop/errors/5c8f662c1823c1001a6d1e86?event_id=5cc5ec1e0038d8cf4df30000&i=gh&m=ci)

## Stacktrace

    src/ui/views/onboarding/seedStore/Ledger.js:187 - Function.e [as mapToProps]

[View full stacktrace](https://app.bugsnag.com/iota-foundation/trinity-desktop/errors/5c8f662c1823c1001a6d1e86?event_id=5cc5ec1e0038d8cf4df30000&i=gh&m=ci)

*Created by Rajiv Shah via Bugsnag*

cvarley100:
Closing due to age and stack trace returning 404

Issue #1386, opened 2019-04-24
Incorrect? Already spent from address error


sloganking:
### Bug description
- Trinity reports unspent address as spent when re-attaching transaction.
- Trinity attempts to "retry" this transaction forever, even after confirmation.

### Possibly related issue
#1036 

### Environment Specs
What device are you running Trinity on?

- Manufacturer: Samsung
- Model: Galaxy S8+
- OS version: Android 9

What build of Trinity are you running? Please provide the full version (you can find it in Settings > About Trinity).

- Trinity version: 0.6.2 (55)

### Steps To Reproduce
1. get network error when generating receiving address but still receive address
2. send funds to generated address from same account
3. Error

### Expected behavior
The wallet should not tell the user that an unspent address is already spent

### Actual behavior
Trinity tells users that the sending address has already been sent from, yet broadcasts the transaction in question (possibly unsafe)

### Errors
![1](https://i.imgur.com/w3LzWAH.png)
![2](https://i.imgur.com/Ca9ftBc.png)
![3](https://i.imgur.com/bwGZ9u8.png)
![4](https://i.imgur.com/vZyvwRA.png)
![5](https://i.imgur.com/RPUZ3X2.png)

sloganking:
Still ongoing in Trinity V 0.7.0 (59)

laumair:
@SleepingFox88 I cannot seem to reproduce this issue. 

What notification did you see after you retried the failed transaction for the first time in history?

sloganking:
Has this issue been fixed? 

I see #1607 marked this issue as a duplicate and is now closed.
Also #1644 marked this issue as closed and has been merged.

I updated mobile trinity to v0.7.2(67) and after an initial "Address already spend from" error message, the problem seems to be fixed for me.

PR #1837, opened 2019-06-20
Desktop: Prepare 0.6.2 release

### LINKED ISSUE #1538: Already spent from address after sending failed
### Bug description
I tried to make a transaction. Internet connection was bad and it failed at `Preparing inputs`. The log reported something like `could not get quorum subtangle milestone`. Directly after that I tried it a second time and this time it took more than a minute to `prepare inputs` and it failed at `Broadcasting` this time. The error log shows the same error again that occurred on the first try. Now this tx is shown in the history and Trinity tries to reattach but always fails with `Already spent from address`. 
After some time my tx got confirmed but after restarting Trinity it's still shown as not confirmed. After refreshing history it's shown as confirmed again but Trinity still tries and fails to reattach.

Seems to be related to: #1036 and maybe #1386

### Environment Specs
What device are you running Trinity on?

- Model: iPhone 8
- OS version: 12.1.1
- Trinity version: 0.7.0(61)

### Steps To Reproduce
1. make a transaction
2. after it failed try again
3. reattach it in history after it failed a second time

### Expected behaviour
Trinity should reattach the failed tx normally.

### Actual behaviour
It always reports `Already spent from address` when trying to reattach the failed transaction.

### Errors
IMG_4587
IMG_E4588
IMG_E4590

cvarley100:
Fix: Migration not successful when migrating from 0.4.x and below to latest (#1832)
Fix: Incorrect already spent from address errors on transaction retry (#1835)
Fix: Incorrect transaction failure alert when successfully broadcast (#1835)
Fix: Quorum being conducted on transaction account syncs when explicitly turned off (#1835)
Fix: Error-related crashes (#1833)
Update: Add remote node list endpoint back-ups (#1811)
Update: Add more verbose error log messages (#1834)
Update: New translations (#1826)

Closes #1538, #1386, #1815, #1422

### DIFF
--- a/src/desktop/src/ui/global/About.js
+++ b/src/desktop/src/ui/global/About.js
@@ -59,6 +59,17 @@ export class AboutComponent extends React.PureComponent {
 
                     <article>
                         <Scrollbar>
+                            <h5>0.6.2</h5>
+                            <ul>
+                                <li>Fix: Migration not successful when migrating from 0.4x and below to latest</li>
+                                <li>Fix: Incorrect already spent from address errors on transaction retry</li>
+                                <li>Fix: Incorrect transaction failure alert when successfully broadcast</li>
+                                <li>Fix: Quorum being conducted on transaction account syncs when explicitly turned off</li>
+                                <li>Fix: Error-related crashes</li>
+                                <li>Fix: Handle exception if password salt is missing from keychain</li>
+                                <li>Update: Add remote node list endpoint back-ups</li>
+                                <li>Update: New translations</li>
+                            </ul>
                             <h5>0.6.1</h5>
                             <ul>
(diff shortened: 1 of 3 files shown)


src/desktop/src/ui/global/About.js
```diff
@@ -59,6 +59,17 @@ export class AboutComponent extends React.PureComponent {
 
                     <article>
                         <Scrollbar>
+                            <h5>0.6.2</h5>
+                            <ul>
+                                <li>Fix: Migration not successful when migrating from 0.4x and below to latest</li>
```
rihardsgravis:
```suggestion
                                <li>Fix: Migration not successful when migrating from 0.4.x and below to latest</li>
```

rihardsgravis (approved):

PR #1884, opened 2019-06-28
Specify maxLength for message textboxes

### LINKED ISSUE #1879: App will crash if message is too long when creating a QR code
## Error in Trinity Desktop

**Error** in **/wallet/receive**
code length overflow. (23668>23648)

[View on Bugsnag](https://app.bugsnag.com/iota-foundation/trinity-desktop/errors/5d13e5004454840019921aa4?event_id=5d13e7b80045fd3649fd0000&i=gh&m=ci)

## Stacktrace

    iota://dist/bundle.js:32639 - Function.QRCode.createData

[View full stacktrace](https://app.bugsnag.com/iota-foundation/trinity-desktop/errors/5d13e5004454840019921aa4?event_id=5d13e7b80045fd3649fd0000&i=gh&m=ci)

*Created by Rajiv Shah via Bugsnag*

Thoralf-M:
Fixes #1879

## Type of change


src/mobile/src/ui/components/MultiTextInput.js
```diff
@@ -9,6 +9,8 @@ import { width, height } from 'libs/dimensions';
 import { Styling } from 'ui/theme/general';
 import { getThemeFromState } from 'shared-modules/selectors/global';
 import AmountTextInput from './AmountTextInput';
+import { /*MAX_TAG_LENGTH,*/ MAX_MESSAGE_LENGTH } from 'shared-modules/libs/iota/utils';
```
rajivshah3, line 11: ESLint is showing an error here:

```

/workdir/src/mobile/src/ui/components/MultiTextInput.js
--
  | 12:56  error  Absolute imports should come before relative imports  import/first

```
This can be solved by moving this line above line 11

rajivshah3 (changes requested):

src/desktop/src/ui/components/input/Text.js
```diff
@@ -47,6 +48,7 @@ export default class Text extends React.PureComponent {
                         type="text"
                         value={value}
                         onChange={(e) => onChange(e.target.value)}
+                        maxLength={MAX_MESSAGE_LENGTH}
```
rajivshah3: This will limit all text fields to 1093 characters. Instead, we should pass a `maxLength` prop which will be used similar to the `label` prop

Thoralf-M: I'm not sure how that should look, should it be like
```
...   
/** Limit max text length */
maxLength: PropTypes.integer,
...
const { value, label, maxLength, onChange, disabled } = this.props;
```
?

rajivshah3: Yep, that's correct. You'll still need this part though:

```
<input
    maxLength // this is short for maxLength={maxLength}
/>
```

src/desktop/src/ui/components/input/Text.js
```diff
@@ -20,6 +20,8 @@ export default class Text extends React.PureComponent {
          * @param {string} value - Current input value
          */
         onChange: PropTypes.func.isRequired,
+        /** Limit max text length */
+        maxLength: PropTypes.integer,
```
rajivshah3: Whoops, this is actually called `number` 😬 

```suggestion
        maxLength: PropTypes.number,
```

rajivshah3 (approved):
Looks good, thanks @Thoralf-M !

Tested on iOS and macOS

PR #1887, opened 2019-06-29
Desktop: Change price quoting

### LINKED ISSUE #1885: Price quoting in Trinity  is the wrong way around.
In the pricing section on the Trinity Desktop it says BTC/iota but the price is in Satoshis. So as it is displayed  right now BTC is the base currency and Iota is the quoting currency but it’s quoted in Satoshis. So if we want to continue to price it in sats (which I guess we all want since its the most common form) it should read Iota/btc. Same goes for eth and USD.  For example: USD/IOTA means how many iotas do I need to get 1$ (answer:2.44) but we want the other way around 1 IOTA is worth 0.41$ that’s whhy it should read IOTA/USD

Thoralf-M:
Fixes #1885 for more information read #1885 

## Type of change


rajivshah3 (approved):
Looks good, thanks @Thoralf-M !

Tested on macOS

PR #2222, opened 2019-09-16
Shared: Fix auth headers

cvarley100:
## Description

- Replaces `token` with correct header `username` when changing to/adding a basic auth node

## Type of change

# How Has This Been Tested?

Tested on Mac and iOS

# Checklist:


rihardsgravis (approved):

src/shared/storage/index.js
```diff
@@ -264,7 +264,7 @@ class Node {
                 custom: true,
                 pow,
                 password: node.password,
-                token: node.token,
+                username: node.username,
```
laumair, line 267: I believe this change will require a realm schema migration.

PR #2385, opened 2019-12-08
Shared: Add CDA parsing via QR and Paste

cvarley100:
# Description 

Parse CDA links scanned via QR and paste

## Type of change

# How Has This Been Tested?

Tested on iOS

# Checklist:


src/shared/actions/alerts.js
```diff
@@ -47,8 +47,8 @@ const dismiss = () => ({ type: AlertsActionTypes.HIDE });
  *
  * @returns {function} dispatch
  */
-export const generateAlert = (category, title, message, closeInterval, err) => (dispatch) => {
-    dispatch(generate(category, title, message, closeInterval));
+export const generateAlert = (category, title, message, closeInterval, err, timeout) => (dispatch) => {
+    setTimeout(() => dispatch(generate(category, title, message, closeInterval)), timeout);
```
rajivshah3:
```suggestion
    if (!timeout) {
        dispatch(generate(category, title, message, closeInterval));
    } else {
        setTimeout(() => dispatch(generate(category, title, message, closeInterval)), timeout);
    }
```

If `generateAlert` is called and `timeout` is not specified, a timer will be created with a timeout of 0. On mobile, this actually results in a [call](https://github.com/facebook/react-native/blob/c20070f10458d48d6ac1eaac49e681e932bfb9fd/Libraries/Core/Timers/JSTimers.js#L259) to the native timing module. Even though the native timing module will immediately call the callback if `timeout` is 0 (see [iOS](https://github.com/facebook/react-native/blob/c20070f10458d48d6ac1eaac49e681e932bfb9fd/React/Modules/RCTTiming.m#L280-L284) and [Android](https://github.com/facebook/react-native/blob/c20070f10458d48d6ac1eaac49e681e932bfb9fd/ReactAndroid/src/main/java/com/facebook/react/modules/core/Timing.java#L347-L352)), using `setTimeout` for something that will run immediately is still unnecessary IMHO. This change will also resolve most of the failing tests for `shared`.

src/shared/libs/iota/utils.js
```diff
@@ -546,3 +546,52 @@ export const withRequestTimeoutsHandler = (timeout) => {
...
+        message: undefined,
+        expectedAmount: undefined,
+        timeoutAt: undefined,
+    };
+
+    try {
+        const url = new URL(input, true)
+        const { address, message, amount, timeoutAt } = url.query
+
+        if (address.match(/^[A-Z9]{90}$/)) {
```
rajivshah3:
```suggestion
        if (address.match(VALID_ADDRESS_WITH_CHECKSUM_REGEX)) {
```

src/shared/libs/iota/utils.js
```diff
@@ -546,3 +546,52 @@ export const withRequestTimeoutsHandler = (timeout) => {
...
+            result.address = address.toUpperCase();
```
rajivshah3:
```suggestion
            result.address = address;
```

If the address matches the above regex, shouldn't it already be upper case?

rajivshah3 (changes requested):

src/desktop/src/ui/global/Alerts.js
```diff
@@ -47,7 +47,7 @@ export class AlertsComponent extends React.PureComponent {
         if (nextProps.alerts.category && nextProps.alerts.category.length && nextProps.alerts.closeInterval > 0) {
             this.timeout = setTimeout(() => {
                 this.props.dismissAlert();
-            }, nextProps.alerts.closeInterval);
+            }, nextProps.alerts.closeInterval || 0);
```
rajivshah3, line 50:
```suggestion
            }, nextProps.alerts.closeInterval);
```
If `closeInterval > 0` (in the `if` statement), shouldn't `closeInterval || 0` always evaluate to `closeInterval`?

src/shared/actions/alerts.js
cvarley100: Okay good catch.

src/desktop/src/ui/global/Alerts.js
cvarley100: Could be undefined

src/shared/actions/transfers.js
```diff
@@ -1022,3 +1025,46 @@ export const retryFailedTransaction = (accountName, bundleHash, seedStore, quoru
...
+          dispatch(setCDAContent(data));
+          dispatch(setSendAddressField(data.address));
+          if (data.expectedAmount) {
+              dispatch(setSendAmountField(data.expectedAmount.toString()));
+          }
+          if (data.message) {
+              dispatch(setSendMessageField(data.message));
+          }
+      } catch (err) {
+          if (err.message === 'Expired timeout.') {
```
laumair: `Expired timeout.` should probably go to https://github.com/iotaledger/trinity-wallet/blob/feat/exchanges/src/shared/libs/errors.js

src/desktop/src/ui/views/wallet/Dashboard.js
```diff
@@ -67,6 +79,16 @@ class Dashboard extends React.PureComponent {
         );
     };
 
+    onPaste = (e) => {
+        const data = (event.clipboardData || window.clipboardData).getData('text')
+        const parsedCDALink = parseCDALink(data);
+        if (parsedCDALink){
+            this.props.verifyCDAContent(parsedCDALink);
+            this.props.history.push('/wallet/send');
+            return e.preventDefault();
```
laumair, line 88: Shouldn't we _always_ prevent the default event? As in shouldn't `e.preventDefault` always be called when `onPaste` callback is triggered?

cvarley100: No, because we want it to fall through to the regular paste function(s) otherwise.

laumair (approved):

PR #2502, opened 2020-01-20
fix/mobile: autocapitalise seed entry

### LINKED ISSUE #2501: Seed should be autocapitalised
### Bug description
Users must turn on caps lock to manually enter a seed. It ought to be autocapitalised to avoid this friction point.

### Environment Specs

iOS/Android 1.2.0

cvarley100:
# Description

Autocapitalises seed on entry.

Fixes #2501 

## Type of change

# How Has This Been Tested?

Tested on Android and iOS

# Checklist:

### DIFF
--- a/src/mobile/src/ui/components/CustomTextInput.js
+++ b/src/mobile/src/ui/components/CustomTextInput.js
@@ -226,10 +226,11 @@ class CustomTextInput extends Component {
     onChangeText(value) {
         const { isPasswordInput, isSeedInput, onValidTextChange } = this.props;
         if (isSeedInput) {
-            if (value && !value.match(VALID_SEED_REGEX)) {
+            const valueCapitalized = value.toUpperCase()
+            if (valueCapitalized && !valueCapitalized.match(VALID_SEED_REGEX)) {
                 return;
             }
-            return onValidTextChange(trytesToTrits(value));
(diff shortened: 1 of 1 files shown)


laumair (approved):
Looks good to me!

src/mobile/src/ui/components/CustomTextInput.js
```diff
@@ -226,10 +226,11 @@ class CustomTextInput extends Component {
     onChangeText(value) {
         const { isPasswordInput, isSeedInput, onValidTextChange } = this.props;
         if (isSeedInput) {
-            if (value && !value.match(VALID_SEED_REGEX)) {
+            const valueCapitalized = value.toUpperCase()
+            if (valueCapitalized && !valueCapitalized.match(VALID_SEED_REGEX)) {
```
holiman, line 230: Both before this change and after -- so if `valueCapitalized` is `""`, the regex-check won't trigger, and the empty seed will go to `onValidTextChange` -- ?

holiman: Should probably be
```javascript
if (!valueCapitalized || !valueCapitalized.match(VALID_SEED_REGEX)) {
    return;
}
```

PR #2692, opened 2020-04-03
Increase AttachToTangle timeout

svenger87:
# Description of change

Hornet remote POW (AVX enabled) on a hetzner CX21 instance often fails because the pow takes slightly more then 25 seconds.
Therefore Trinity falls back to local POW.

## Type of change

## How the change has been tested

Compiled Trinity from souce with increased timeout in config.js

API request performed several times with measured duration of POW:

DURATION: 22350
DURATION: 14593
DURATION: 8931
DURATION: 14619
DURATION: 26030 < Boing...this one would have failed in a trinity tx
DURATION: 19363
DURATION: 19687
DURATION: 10816
DURATION: 28896 < Boing...this one would have failed in a trinity tx

Tested with this script: https://cdn.discordapp.com/attachments/630794683920154676/694998985546661919/api_attachToTangle.sh

## Change checklist

Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [x] My code follows the contribution guidelines for this project


rajivshah3 (approved):

rajivshah3:
While I think this change is fine, I did look into the PoW implementation that Hornet uses because >25 sec for AVX PoW on an average size bundle seems unusually long. I think iota.go (and thus Hornet) would benefit from using the optimized PCurl implementation from [iotaledger/iota_common](https://github.com/iotaledger/iota_common/blob/master/common/crypto/curl-p/README.md).

svenger87:
I´d be happy if there are better options. @muXxer

svenger87:
In my manual build i increased the timeout to 60000. It seems trinity is NOT waiting 60s for remote POW. Am i missing something?

laumair:
@svenger87 

> Hornet remote POW (AVX enabled) on a hetzner CX21 instance often fails because the pow takes slightly more then 25 seconds.
Therefore Trinity falls back to local POW.

Trinity should not instantly fall to local proof-of-work because of this [handler](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/libs/iota/utils.js#L529) used in the [attachToTangle](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/libs/iota/extendedApi.js#L597) api. 

> It seems trinity is NOT waiting 60s for remote POW.

Does it error out or just retries? If it retries, then its intentional.

svenger87:
@laumair Remote POW gets a request timeout and falls back to local POW. 
Which is particularly good. But I was thinking giving the node a bit more time to complete POW might make sense. 

Feel free to try a few transactions with https://hornet.spdns.org:443

cvarley100 (approved):

PR #2957, opened 2020-08-14
Update promotion & reattachment implementation

laumair:
# Description of change

- Updated promotion & reattachment implementation
- Added a fallback mechanism to promotion & reattachment (if `getTipInfo` is unavailable)
- Removed the option for users to manually promote a transaction
- Removed the setting for auto-promotion (it will be enabled for all users)

## Type of change

Choose a type of change, and delete any options that are not relevant.

## How the change has been tested

Manually tested with comnet nodes

## Change checklist

Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [x] My code follows the contribution guidelines for this project


src/shared/libs/iota/extendedApi.js
```diff
@@ -520,6 +513,37 @@ const checkAttachToTangleAsync = (node) => {
...
+        body: JSON.stringify({
+            command: 'getTipInfo',
+            tailTransaction: tailTransactionHash,
+        }),
+        headers: new Headers({
+            'Content-Type': 'application/json',
+            'X-IOTA-API-Version': IRI_API_VERSION,
+        }),
+    }).then((response) => {
+        return response.json().then((res) => {
```
rajivshah3, line 537: Should we check `response.ok` first before parsing the JSON? We did that above for `checkAttachToTangleAsync`

laumair: We are extracting error message from the parsed object `throw new Error(res.error);`, so I guess in this case it is fine to parse it before.

rajivshah3: Ok. The reason I ask is because if there's JSON parsing error, I don't think it'll be caught because the `throw` is inside `then`, but `response.json()` will run (and throw) before the `then` block

cvarley100 (approved):

